PR #14, opened 2014-11-28
Upgrade to JUnit 4.11

nekop:
- It doesn't break existing JUnit 3 tests, all tests run fine
- Allows us to run single test like "mvn test -Dtest=ThreadPoolTestCase#testQueuelessKeepAlive"


dmlloyd:
Resolved via #30 and #31.

PR #16, opened 2015-09-21
[WFLY-5332]: New ThreadPool Executor reusing Idle Threads

panossot:
WFLY-5332 : https://issues.jboss.org/browse/WFLY-5332

Creating a new Thread Pool Executor that will be used to create an Unbound Queue Thread Pool that reuses idle threads in the Application Server.

### DIFF
--- a/src/main/java/org/jboss/threads/JBossRejectedExecutionHandler.java
+++ b/src/main/java/org/jboss/threads/JBossRejectedExecutionHandler.java
@@ -1,0 +1,16 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.jboss.threads;
+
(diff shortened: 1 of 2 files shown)


wolfc:
What's the difference with a regular ThreadPoolExecutor?
Clearly this is a fork of some other code. It would make more sense to split it into 2 commits. The original, followed by the changes.

panossot:
Carlo you are right. I will split it in two commits.

panossot:
The differences are visible in this commit : https://github.com/panossot/jboss-threads/commit/b44a7731c4672b986c9133f8b28ad0bbf8375768

wolfc:
Now you are making a mess of things. The PR should only contain 2 commits.

panossot:
I will create another one with only two commits.

panossot:
New PR with two commits : https://github.com/jbossas/jboss-threads/pull/17
Closing this PR...

PR #17, opened 2015-09-23
[WFLY-5332] : Modified JBoss ThreadPool Executor reusing Idle Threads 

panossot:
[WFLY-5332] : https://issues.jboss.org/browse/WFLY-5332

Modifying JBoss Thread Pool Executor, that is used to create an Unbound Queue Thread Pool in the Application Server, to reuse idle threads.


panossot:
Replaced by https://github.com/jbossas/jboss-threads/pull/19

PR #19, opened 2016-02-12
JBTHR-34 : Allow idle thread reusage.

panossot:
[JBTHR-34] : https://issues.jboss.org/browse/JBTHR-34

### DIFF
--- a/src/main/java/org/jboss/threads/JBossRejectedExecutionHandler.java
+++ b/src/main/java/org/jboss/threads/JBossRejectedExecutionHandler.java
@@ -1,0 +1,16 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.jboss.threads;
+
+import java.util.concurrent.RejectedExecutionHandler;
+
+/**
+ *
+ * @author panos
+ */
+public interface JBossRejectedExecutionHandler {
+    void rejectedExecution(Runnable r, JBossThreadPoolExecutorReuseIdleThreads executor);
+}
--- a/src/main/java/org/jboss/threads/JBossThreadPoolExecutorReuseIdleThreads.java
+++ b/src/main/java/org/jboss/threads/JBossThreadPoolExecutorReuseIdleThreads.java
@@ -1,0 +1,1878 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2015, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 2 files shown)


dmlloyd:
I still have not thoroughly analyzed this yet.  This is clearly a fork of OpenJDK's ThreadPoolExecutor.  What is the original version this was copied from (for comparison purposes)?  I'd like to note that the original ThreadPoolExecutor is sufficiently complex that even after the analysis already done on it, problems crop up from time to time.  Various people (I, and before me, Adrian Brock, and possibly others) explored similar changes based on the original TPE code and have discovered race conditions in the past, one of which in particular can cause the thread pool task queue to contain tasks while there are zero running threads.  This is the basis of my overall skepticism, if you were wondering about that.

Legally speaking: we don't have the copyright on this.  I don't think we can legally relicense the code under LGPL or ASL or anything else; you'll have to check with the legal team on that.

panossot:
Hello David. For now, I have forked the code of OpenJDK's ThreadPoolExecutor, yes, for comparison purposes. (I can try to extend the original class later on). I understand that it touches core functionalities and a change cannot be tested sufficiently. The problem with the queue containing tasks while there are no running threads, I cannot imagine the reason (For this point I would need a reproducer. Does it happen with the allow-core-timeout set to true? They should be processed with the next incoming task, though,  that will cause the addition of a new Worker.).  Regards.

dmlloyd:
Yes I believe that particular case had allow-core-timeout set to true, which let threads expire at the same moment that a task went into the queue.

dmlloyd:
I have some future plans (hopefully for EAP 7.2) for reworking some of this thread pool stuff, as follows:

1. Develop a new thread pool type (working name is HighScaleThreadPoolExecutor) which supports thread pool minimum and maximum sizesup to 16K threads, striped work distribution, and better algorithms which I am developing for detecting thread pool busyness (including a customized special push-pull queue type that can hold waiting threads or waiting tasks).  This thread pool will support all the same run-time metrics and tunables as ThreadPoolExecutor but hopefully will both scale better and dispatch more efficiently.  Specifically it gets rid of the broken "don't go beyond core threads unless the queue is full" mentality which makes using unbounded queues be problematic with core and max sizes.
2. Deprecate and ultimately remove all of the other custom thread pool implementations in the jboss-threads project, specifically JBossThreadPoolExecutor, QueueExecutor, QueuelessExecutor.
3. Eliminate the threads subsystem, or at the very least remap all of the thread pool types to the new thread pool.

panossot:
If you think that the modifications of this PR could be useful, I would be happy to help.

panossot:
If I understand correctly, there will be one more queue for the idle/waiting threads and when a new task is received, before going inside the task queue, it would check the waiting threads queue, right?

dmlloyd:
No, there will be a single, special-purpose queue which can handle both tasks and threads in an atomic action.

panossot:
Interesting. Looking forward to seeing it.

PR #24, opened 2018-01-30
[WFLY-9742] Clear the TCCL to avoid classloader leaks

markusdlugi:
As described in [WFLY-9742](https://issues.jboss.org/browse/WFLY-9742), this change ensures that any JBossThreads created by the JBossThreadFactory will not leak classloaders because of using the TCCL inherited by its parent thread.

### DIFF
--- a/src/main/java/org/jboss/threads/JBossThreadFactory.java
+++ b/src/main/java/org/jboss/threads/JBossThreadFactory.java
@@ -111,6 +111,10 @@ public final class JBossThreadFactory implements ThreadFactory {
         }
         thread.setThreadNameInfo(nameInfo);
         thread.setName(nameInfo.format(thread, namePattern));
+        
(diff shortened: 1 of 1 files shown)


markusdlugi:
@dmlloyd Is the PR okay this way? Any change requests?

dmlloyd:
The change itself looks OK, however I do need a corresponding JIRA from https://issues.jboss.org/browse/JBTHR and it should be referenced in the PR and commit, and probably the code as well (instead of the WFLY- issue).  The JBTHR issue may be linked to the WFLY issue.  Also one more thing...

dmlloyd:
We require that all contributions be made under the terms of the MIT License, http://www.opensource.org/licenses/mit-license.php. Do you agree to the these terms?

dmlloyd:
Resolved via #32 and #33 .

PR #25, opened 2018-02-07
[JBTHR-49] : Current Thread Count  should not be less than Core Thread Count.

panossot:
[JBTHR-49] : https://issues.jboss.org/browse/JBTHR-49


dmlloyd:
Merged via #28.

PR #29, opened 2018-02-19
JBTHR-50 Threads can end up hung when they move to GAVE_UP state

stuartwdouglas:
Once a thread has moved to GAVE_UP it is not valid to move it
back to WAITING as it may have already been discarded from the
queue


dmlloyd:
Merged in #30 and #31.

PR #39, opened 2018-04-25
[JBTHR-57] Tests for EnhancedQueueExecutor keepalive setting

LittleJohnII:
Introducing a few simple tests for the `EnhancedQueueExecutor`, centered mainly on the `keepalive` setting. All tests pass.

These are a part of the test development for EAP7-488 and EAP7-716.

When you have the time, let me know what you think. Thanks.


dmlloyd:
This needs a JBTHR JIRA (and commit & PR title edited accordingly).

Also, in general we should only use sleep in tests when _necessary_.  This particular test only adds about 4 seconds to the build so it's OK in this instance, but in the future please be aware that test execution time is important as well.

PR #46, opened 2019-01-02
JBTHR-70 Only log the version at DEBUG level, not INFO

gsmet:
* https://issues.jboss.org/browse/JBTHR-70


gsmet:
OK, closing after discussion with @dmlloyd : we need to be more clever than that and have a way to enable version logging globally.

PR #51, opened 2019-08-15
[JBTHR-77] Add new tests for testing EnhacedThreadPoolExecutor

mnovak1:


mnovak1:
@dmlloyd hi David, could you or some one else review this PR, please?

dmlloyd (approved):
I'll merge this, with the caveat that if any false negatives appear due to timing problems, that test will be disabled until the scenario is definitively addressed.

PR #52, opened 2019-08-30
[JBTHR-77] Add new tests for testing EnhacedThreadPoolExecutor

mnovak1:


dmlloyd (approved):
Merging with the same caveat as I described in https://github.com/jbossas/jboss-threads/pull/51#pullrequestreview-490673496 .

PR #57, opened 2020-01-31
JBTHR-71: Improve EnhancedQueueExecutor lock locality

carterkozak:
Previously the headLock and tailLock were accessed across
padding fields causing cache misses and regressed performance
compared to execution with locks disabled.


dmlloyd:
Thanks @carterkozak!  One detail that has to be worked out first:

We require that all contributions be made under the terms of the MIT License, http://www.opensource.org/licenses/mit-license.php. Do you agree to the these terms?

(If you have already agreed on a previous PR, please link to that PR.)

carterkozak:
> We require that all contributions be made under the terms of the MIT License, http://www.opensource.org/licenses/mit-license.php. Do you agree to the these terms?

I do.

dmlloyd:
Great! Thanks for the contribution!

carterkozak:
Thanks for your help and the quick review!

PR #59, opened 2020-02-03
JBTHR-80: EnhancedQueueExecutor avoids allocation in tight loops

carterkozak:
When allocation is required, it occurs at most once, and objects
are reused.

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -1509,6 +1509,7 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
         private QNode getOrAddNode() {
             TaskNode head;
             QNode headNext;
+            PoolThreadNode newNode = null;
             for (;;) {
                 head = EnhancedQueueExecutor.this.head;
                 headNext = head.getNext();
@@ -1519,8 +1520,10 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
                         return taskNode;
                     }
                 } else if (headNext instanceof PoolThreadNode || headNext == null) {
-                    PoolThreadNode newNode;
-                    newNode = new PoolThreadNode((PoolThreadNode) headNext, Thread.currentThread());
+                    if (newNode == null) {
+                        newNode = new PoolThreadNode(Thread.currentThread());
+                    }
+                    newNode.setNext(headNext);
                     if (head.compareAndSetNext(headNext, newNode)) {
                         return newNode;
                     }
@@ -1696,6 +1699,7 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
         QNode tailNext;
         TaskNode tail = this.tail;
         final int result;
+        TaskNode node = null;
         for (;;) {
             tailNext = tail.getNext();
             if (tailNext instanceof TaskNode) {
@@ -1779,7 +1783,10 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
                     break;
                 }
                 // queue size increased successfully; we can add to the list
-                TaskNode node = new TaskNode(runnable);
+                if (node == null) {
+                    // avoid re-allocating TaskNode instances on subsequent iterations
+                    node = new TaskNode(runnable);
+                }
                 // state change ex5:
                 //   tail(snapshot).next ← new task node
                 // cannot succeed: sh2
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1519,8 +1520,10 @@ private QNode getOrAddNode() {
                         return taskNode;
                     }
                 } else if (headNext instanceof PoolThreadNode || headNext == null) {
-                    PoolThreadNode newNode;
-                    newNode = new PoolThreadNode((PoolThreadNode) headNext, Thread.currentThread());
+                    if (newNode == null) {
+                        newNode = new PoolThreadNode(null, Thread.currentThread());
```
dmlloyd: This constructor is only called here IIRC, so the first parameter can be safely dropped instead of always being `null`.

carterkozak: Good call. It turns out this was the only constructor for any subtype of QNode that used the QNode(QNode next) ctor with a non-null value, so I've removed that entirely as well.

carterkozak:
Do you think this is sufficient to switch the default value of head-lock and tail-lock to false? I haven't been able to produce a scenario where the locks improve overall performance.

dmlloyd:
I would need more data to be certain.  Would it be possible to test with a constrained CPU set (e.g. 1, 2, 4 virtual CPUs)?  I would normally use `taskset` for this on Linux.

carterkozak:
Undertow server, 8 I/O threads, 32 concurrent client requests (client request duration is the measured portion):

```
no locks:

1 core:
Benchmark                           (httpServer)   Mode  Cnt     Score      Error  Units
Benchmarks.parallelRequestsGet  UNDERTOW_HANDLER  thrpt    3  6522.986 ± 6620.209  ops/

2 cores:
Benchmark                           (httpServer)   Mode  Cnt      Score       Error  Units
Benchmarks.parallelRequestsGet  UNDERTOW_HANDLER  thrpt    3  14403.326 ± 14446.680  ops/s

4 cores:
Benchmark                           (httpServer)   Mode  Cnt      Score      Error  Units
Benchmarks.parallelRequestsGet  UNDERTOW_HANDLER  thrpt    3  52410.909 ± 4414.995  ops/s

locks:

1 core:
Benchmark                           (httpServer)   Mode  Cnt     Score      Error  Units
Benchmarks.parallelRequestsGet  UNDERTOW_HANDLER  thrpt    3  6580.091 ± 8298.738  ops/s

2 cores:
Benchmark                           (httpServer)   Mode  Cnt      Score       Error  Units
Benchmarks.parallelRequestsGet  UNDERTOW_HANDLER  thrpt    3  23955.632 ± 77759.587  ops/s

4 cores:
Benchmark                           (httpServer)   Mode  Cnt      Score       Error  Units
Benchmarks.parallelRequestsGet  UNDERTOW_HANDLER  thrpt    3  44196.603 ± 38596.792  ops/s

synchronized locking:

1 core:
Benchmark                           (httpServer)   Mode  Cnt     Score      Error  Units
Benchmarks.parallelRequestsGet  UNDERTOW_HANDLER  thrpt    3  6351.627 ± 5505.567  ops/s

2 cores:
Benchmark                           (httpServer)   Mode  Cnt      Score        Error  Units
Benchmarks.parallelRequestsGet  UNDERTOW_HANDLER  thrpt    3  22115.672 ± 101593.181  ops/s

4 cores
Benchmark                           (httpServer)   Mode  Cnt      Score       Error  Units
Benchmarks.parallelRequestsGet  UNDERTOW_HANDLER  thrpt    3  46051.980 ± 14915.686  ops/s
```

The two-core case is interesting, we gain a lot of performance by adding locks, from 14k ops/sec to 23k ops/sec, though the variance also increases from 14k to 77k or 100k depending on the lock type ( running the tests for longer may provide more clarity around the variance).
I think the locks are helpful when many threads are submitting tasks on an oversubscribed host, but hold us back as the core count increases to meet the need.
For a server where most work is submitted form a small I/O task pool there's less contention than one might expect from other use cases (or servlet async dispatch where tasks are submitted from the task pool).
The test runs without locking appeared to spin up faster than those with locks.

dmlloyd:
The error margin is larger than the score in some cases...

carterkozak:
Lots of threads on very few cores. It might be a better test if I reduced the requesting threads from 32 as well, but that may not produce as much contention either.

carterkozak:
@dmlloyd I think this is a better example, I've reproduce the issue without any of the web server/client overhead: https://github.com/carterkozak/jboss-threads-performance

I would have contributed the benchmark directly to jboss-threads, but I'm not as familiar with maven as I should be.
The jmh benchmark shells out to `taskset` to test various numbers of cores, though the single core tests won't be able to take advantage of the spinlock optimization for a single core.

carterkozak:
I've added a few more results to the readme in the linked repository. Gist is running without locks always appears to outperform locks, the spin/reentrant locks have trouble past ~200,000 ops/sec.
Using synchronized instead of lock/unlock scales better, but doesn't come close to running without locks.
For comparison, a standard fixed size thread pool executor can handle about 700,000 ops/second, it's outperformed by EnhancedQueueExecutor when locks are disabled, but not when locks or synchronization are used.

PR #61, opened 2020-02-20
Implement synchronized lock mode

carterkozak:
This change adds two new properties:
`jboss.threads.eqe.head-synchronized` and
`jboss.threads.eqe.tail-synchronized` which both default to `false`. The
default configuration is not modified.

Benchmark results for 32 threads submitting tasks to a pool with 100
threads on a 14 core system with SMT:
```
Benchmark                           (cores)             (factory)   Mode  Cnt        Score       Error  Units
ExecutorBenchmarks.benchmarkSubmit       28  THREAD_POOL_EXECUTOR  thrpt    4   783474.478 ± 15490.365  ops/s
ExecutorBenchmarks.benchmarkSubmit       28          EQE_NO_LOCKS  thrpt    4  1809063.938 ± 13020.180  ops/s
ExecutorBenchmarks.benchmarkSubmit       28         EQE_SPIN_LOCK  thrpt    4   349376.970 ±  6046.860  ops/s
ExecutorBenchmarks.benchmarkSubmit       28      EQE_SYNCHRONIZED  thrpt    4   530502.587 ± 14570.999  ops/s
ExecutorBenchmarks.benchmarkSubmit       28    EQE_REENTRANT_LOCK  thrpt    4   346651.156 ±  4326.834  ops/s
```

32 threads submitting tasks to fixed executors with 16 threads, given 28
logical cores:
```
Benchmark                           (cores)             (factory)   Mode  Cnt       Score       Error  Units
ExecutorBenchmarks.benchmarkSubmit       28  THREAD_POOL_EXECUTOR  thrpt    4  825319.589 ±  4505.611  ops/s
ExecutorBenchmarks.benchmarkSubmit       28          EQE_NO_LOCKS  thrpt    4  557950.673 ± 17938.335  ops/s
ExecutorBenchmarks.benchmarkSubmit       28         EQE_SPIN_LOCK  thrpt    4  348318.297 ± 10356.888  ops/s
ExecutorBenchmarks.benchmarkSubmit       28      EQE_SYNCHRONIZED  thrpt    4  534380.322 ± 11343.607  ops/s
ExecutorBenchmarks.benchmarkSubmit       28    EQE_REENTRANT_LOCK  thrpt    4  349117.764 ±  7952.673  ops/s
```

28 threads submitting tasks to fixed executors with 14 threads on 4
logical cores:
```
Benchmark                           (cores)             (factory)   Mode  Cnt       Score       Error  Units
[...]

PR #62, opened 2020-02-26
Additional spin loop modifications

dmlloyd:


dmlloyd:
I've actually added a couple things here which I'm going to test out separately.

carterkozak:
I've run benchmarks on this change with the commit from #61

Numbers should be comparable to the results here: https://github.com/jbossas/jboss-threads/pull/61#issuecomment-590093991
This appears to scale much better with cores in my environment, while suffering a bit when only two cores are available.

```
# Run complete. Total time: 00:34:37

Benchmark                           (cores)  (executorThreads)           (factory)   Mode  Cnt        Score        Error  Units
ExecutorBenchmarks.benchmarkSubmit        1              16,16        EQE_NO_LOCKS  thrpt    4   328342.630 ±  48467.042  ops/s
ExecutorBenchmarks.benchmarkSubmit        1              16,16  EQE_REENTRANT_LOCK  thrpt    4   344689.512 ±  10788.572  ops/s
ExecutorBenchmarks.benchmarkSubmit        1              16,16       EQE_SPIN_LOCK  thrpt    4   344274.821 ±  10402.449  ops/s
ExecutorBenchmarks.benchmarkSubmit        1              16,16    EQE_SYNCHRONIZED  thrpt    4   344367.446 ±  10461.115  ops/s
ExecutorBenchmarks.benchmarkSubmit        1            100,200        EQE_NO_LOCKS  thrpt    4   173425.816 ±   7871.344  ops/s
ExecutorBenchmarks.benchmarkSubmit        1            100,200  EQE_REENTRANT_LOCK  thrpt    4   178066.230 ±  15850.644  ops/s
ExecutorBenchmarks.benchmarkSubmit        1            100,200       EQE_SPIN_LOCK  thrpt    4   173207.208 ±  21154.709  ops/s
ExecutorBenchmarks.benchmarkSubmit        1            100,200    EQE_SYNCHRONIZED  thrpt    4   174650.236 ±  20377.631  ops/s
ExecutorBenchmarks.benchmarkSubmit        2              16,16        EQE_NO_LOCKS  thrpt    4    86746.055 ±   2654.920  ops/s
ExecutorBenchmarks.benchmarkSubmit        2              16,16  EQE_REENTRANT_LOCK  thrpt    4   245843.924 ±  22099.366  ops/s
ExecutorBenchmarks.benchmarkSubmit        2              16,16       EQE_SPIN_LOCK  thrpt    4   244212.019 ±  13555.859  ops/s
ExecutorBenchmarks.benchmarkSubmit        2              16,16    EQE_SYNCHRONIZED  thrpt    4   263686.528 ±  85781.158  ops/s
ExecutorBenchmarks.benchmarkSubmit        2            100,200        EQE_NO_LOCKS  thrpt    4   240389.174 ±  17329.255  ops/s
ExecutorBenchmarks.benchmarkSubmit        2            100,200  EQE_REENTRANT_LOCK  thrpt    4   237137.653 ±  23344.433  ops/s
ExecutorBenchmarks.benchmarkSubmit        2            100,200       EQE_SPIN_LOCK  thrpt    4   123044.516 ± 238468.626  ops/s
ExecutorBenchmarks.benchmarkSubmit        2            100,200    EQE_SYNCHRONIZED  thrpt    4   227393.762 ±  40982.082  ops/s
ExecutorBenchmarks.benchmarkSubmit        4              16,16        EQE_NO_LOCKS  thrpt    4   175670.429 ±  11931.894  ops/s
ExecutorBenchmarks.benchmarkSubmit        4              16,16  EQE_REENTRANT_LOCK  thrpt    4   474304.974 ±   6574.753  ops/s
ExecutorBenchmarks.benchmarkSubmit        4              16,16       EQE_SPIN_LOCK  thrpt    4   468244.160 ±  24813.979  ops/s
ExecutorBenchmarks.benchmarkSubmit        4              16,16    EQE_SYNCHRONIZED  thrpt    4   443264.609 ±  35958.383  ops/s
ExecutorBenchmarks.benchmarkSubmit        4            100,200        EQE_NO_LOCKS  thrpt    4   459147.449 ±   2860.134  ops/s
ExecutorBenchmarks.benchmarkSubmit        4            100,200  EQE_REENTRANT_LOCK  thrpt    4   413661.000 ±  10779.054  ops/s
ExecutorBenchmarks.benchmarkSubmit        4            100,200       EQE_SPIN_LOCK  thrpt    4   195625.114 ± 170698.032  ops/s
ExecutorBenchmarks.benchmarkSubmit        4            100,200    EQE_SYNCHRONIZED  thrpt    4   423421.100 ±  44395.811  ops/s
ExecutorBenchmarks.benchmarkSubmit        8              16,16        EQE_NO_LOCKS  thrpt    4   137718.479 ±  15676.465  ops/s
ExecutorBenchmarks.benchmarkSubmit        8              16,16  EQE_REENTRANT_LOCK  thrpt    4   595971.149 ±  10630.429  ops/s
ExecutorBenchmarks.benchmarkSubmit        8              16,16       EQE_SPIN_LOCK  thrpt    4   606321.592 ±  27428.974  ops/s
ExecutorBenchmarks.benchmarkSubmit        8              16,16    EQE_SYNCHRONIZED  thrpt    4   841916.144 ±  28742.039  ops/s
ExecutorBenchmarks.benchmarkSubmit        8            100,200        EQE_NO_LOCKS  thrpt    4   869669.871 ±  78083.842  ops/s
ExecutorBenchmarks.benchmarkSubmit        8            100,200  EQE_REENTRANT_LOCK  thrpt    4   608390.447 ±  29655.185  ops/s
ExecutorBenchmarks.benchmarkSubmit        8            100,200       EQE_SPIN_LOCK  thrpt    4   542612.704 ± 213358.073  ops/s
ExecutorBenchmarks.benchmarkSubmit        8            100,200    EQE_SYNCHRONIZED  thrpt    4   708797.620 ±  86828.079  ops/s
ExecutorBenchmarks.benchmarkSubmit       14              16,16        EQE_NO_LOCKS  thrpt    4   566300.544 ±  80695.246  ops/s
ExecutorBenchmarks.benchmarkSubmit       14              16,16  EQE_REENTRANT_LOCK  thrpt    4   673670.450 ±  19302.221  ops/s
ExecutorBenchmarks.benchmarkSubmit       14              16,16       EQE_SPIN_LOCK  thrpt    4   690270.082 ±  19704.960  ops/s
ExecutorBenchmarks.benchmarkSubmit       14              16,16    EQE_SYNCHRONIZED  thrpt    4  1082418.504 ±  50650.227  ops/s
ExecutorBenchmarks.benchmarkSubmit       14            100,200        EQE_NO_LOCKS  thrpt    4  1416890.535 ± 101921.413  ops/s
ExecutorBenchmarks.benchmarkSubmit       14            100,200  EQE_REENTRANT_LOCK  thrpt    4   675415.013 ±  39911.948  ops/s
ExecutorBenchmarks.benchmarkSubmit       14            100,200       EQE_SPIN_LOCK  thrpt    4   684571.191 ±  41290.408  ops/s
ExecutorBenchmarks.benchmarkSubmit       14            100,200    EQE_SYNCHRONIZED  thrpt    4  1061102.803 ±  33902.446  ops/s
ExecutorBenchmarks.benchmarkSubmit       28              16,16        EQE_NO_LOCKS  thrpt    4  1767051.231 ±  46514.221  ops/s
ExecutorBenchmarks.benchmarkSubmit       28              16,16  EQE_REENTRANT_LOCK  thrpt    4   634645.092 ±  20557.708  ops/s
ExecutorBenchmarks.benchmarkSubmit       28              16,16       EQE_SPIN_LOCK  thrpt    4   636295.110 ±  15038.511  ops/s
ExecutorBenchmarks.benchmarkSubmit       28              16,16    EQE_SYNCHRONIZED  thrpt    4  1059023.747 ±  19376.598  ops/s
ExecutorBenchmarks.benchmarkSubmit       28            100,200        EQE_NO_LOCKS  thrpt    4  1904768.114 ±  99647.988  ops/s
ExecutorBenchmarks.benchmarkSubmit       28            100,200  EQE_REENTRANT_LOCK  thrpt    4   630760.228 ±  18466.746  ops/s
ExecutorBenchmarks.benchmarkSubmit       28            100,200       EQE_SPIN_LOCK  thrpt    4   632733.070 ±   6110.285  ops/s
ExecutorBenchmarks.benchmarkSubmit       28            100,200    EQE_SYNCHRONIZED  thrpt    4  1052828.445 ±  63132.075  ops/s
```

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -2198,12 +2210,14 @@ void park(EnhancedQueueExecutor enhancedQueueExecutor) {
             if (spins > 0) {
                 ThreadLocalRandom tl = ThreadLocalRandom.current();
                 do {
-                    if (tl.nextInt(PARK_SPINS) < YIELD_FACTOR) {
-                        Thread.yield();
-                    }
```
carterkozak: Nice catch

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -2226,12 +2240,14 @@ void park(EnhancedQueueExecutor enhancedQueueExecutor, long nanos) {
...
                     if (unsafe.compareAndSwapInt(this, parkedOffset, STATE_UNPARKED, STATE_NORMAL)) {
                         return;
                     }
+                    if (tl.nextInt(PARK_SPINS) < YIELD_FACTOR) {
+                        Thread.yield();
+                    } else {
+                        JDKSpecific.onSpinWait();
+                    }
```
carterkozak: I wonder how important the `nanoTime` call is in this case as it's ignored while we spin. Perhaps we can round the spin time to zero and park the full duration when PARK_SPINS is reached?

dmlloyd: Well on the one hand, ~128 spins would probably take a few microseconds at most, but on the other hand, `yield` could take an arbitrary amount of time.  Maybe we should only check the time (the second time) if we yielded.

carterkozak: In that case we might as well defer the first nanoTime call until just prior to the first `yield`, that is unless the nanotime call itself has timing implications that make the spin more likely to avoid parking.

dmlloyd: TBH I'm not really sure of the "random yield" approach.  If we yielded, does it make sense to spin again some tiny amount of times afterwards?  Does it make sense to sometimes yield after fewer spins?  I'm suspecting the answer to both of these is "no" - maybe it makes more sense to spin N times, then yield M times, then park.  This would also simplify the elapsed time logic.

carterkozak: Agreed, it's much easier for me to reason about an approach that escalates from spin into yield, and finally park based on well defined inputs.

Separately, though more concerning using the deterministic escalation approach, it's odd that we can yield then exit the loop without a cas check.

dmlloyd: Yeah, that's problematic and I suspect it could result in measurable degradation.  Being deterministic makes it easier to code correctly though I suspect.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1795,6 +1801,12 @@ private int tryExecute(final Runnable runnable) {
                     break;
                 }
                 assert tr == AT_NO;
+                // see if we need to spin
+                if (spins > 0) {
+                    JDKSpecific.onSpinWait();
+                    spins--;
+                    continue;
+                }
```
carterkozak: Actually, does this interact poorly with growth resistance? I think this effectively invalidates non-zero resistance values.

carterkozak:
I've done a bit of testing today and found a sizable (~20%) increase in throughput by unlocking the tail lock prior to a `LockSupport.unpark` invocation (called via `tryExecute`) and acquiring the lock again once unpark completes. This still isn't able to match performance using `synchronized`, but reduces the difference by a third, though I haven't tried restructuring the code to make this optimization with synchronized.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
carterkozak: onSpinWait while holding tailLock probably isn't what we want to do in this case. I'm not confident spinning on submit provides any benefit over additional spins before parking.

dmlloyd: You're right, that (and spinning) is only useful without a lock.  I wonder if it's generally worth unlocking (regardless) before calling `tryAllocateThread`.

dmlloyd:
@carterkozak I've fixed the various issues we've found, and added some more things we discussed including lifting some things from some of your PRs.

dmlloyd:
It looks like something I've done has caused locked mode to really tank in performance.

carterkozak:
Thanks @dmlloyd! On this branch performance is fantastic with the tail lock enabled and head locking disabled. With head and tail locks enabled my profiler shows 78% of samples from `ThreadBody.run` in the `lockHead` method, which accounts for 68% of all samples.

src/main/java/org/jboss/threads/EnhancedQueueExecutorBase1.java
```diff
@@ -78,4 +86,31 @@
...
+
+    // =======================================================
+    // Locks
+    // =======================================================
+
+    final void lockTail() {
+        int spins = 0;
+        for (;;) {
+            if (tailLock == 0) {
+                if (unsafe.compareAndSwapInt(this, tailLockOffset, 0, 1)) {
```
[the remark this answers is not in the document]
dmlloyd: Yeah, can definitely do that.  It should just be cosmetic though...

src/main/java/org/jboss/threads/EnhancedQueueExecutorBase3.java
```diff
@@ -43,16 +46,25 @@
...
     /**
      * Use a spin lock for the head lock.
      */
+    @SuppressWarnings("unused")
     static final boolean HEAD_SPIN = readBooleanPropertyPrefixed("head-spin", true);
 
+    /**
+     * Number of spins before yielding.
+     */
+    static final int YIELD_SPINS = readIntPropertyPrefixed("lock-yield-spins", ProcessorInfo.availableProcessors() == 1 ? 0 : 128);
```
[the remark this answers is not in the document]
dmlloyd: Maybe.  I also thought about scaling based on CPU count, which might provide some useful result.

dmlloyd:
I suspect that the lock surface is still too large in some common scenario, so I'm looking for stupid mistakes first.

dmlloyd:
I think it must be tied to the allocation of the PoolThreadNode.  So as per discussion on #64 I'm going to see if I can combine that class in with ThreadBody.

carterkozak:
I tried to validate that by allocating the node before acquiring the lock, and passing it as an argument. While I wouldn't expect this to provide great performance overall due to unnecessary object allocation for every invocation, it minimizes time spent with the lock, however it didn't make a meaningful difference in my testing (I do think it would be a fantastic optimization even if it's not the long pole here).

Most of the contention in getOrAddNode in my primary scenario (more threads than concurrent tasks) is on head.next CAS, which isn't exclusively protected by the head lock. I wonder if we're better off moving onSpinWait/yield into getOrAddNode instead of using a lock instance? I think getOrAddNode isn't a great deal more expensive than the lock and may allow cycles to be more productive.

dmlloyd:
Yeah I think what we're seeing might be `next` being contended from both head and tail ends... still looking into the minutia.

dmlloyd:
I think your analysis is correct.  Once my current test run completes, I'm going to move the reused node creation to the start of the thread execution, and move the spin logic directly into that method, and drop the head lock completely.

dmlloyd:
I now have essentially equivalent performance between lock and lockless mode.  The most benefit arises from many cores though.  There's probably still more improvement possible on the submit side; maybe by way of padding the queue nodes as described in #64.

dmlloyd:
I've introduced an assertion error.  Checking that out now.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -138,7 +135,7 @@
     /**
      * Update the summary statistics.
      */
-    static final boolean UPDATE_STATISTICS = readBooleanPropertyPrefixed("statistics", true);
+    static final boolean UPDATE_STATISTICS = readBooleanPropertyPrefixed("statistics", false);
```
carterkozak: Was the UPDATE_STATISTICS default value change intentional? I'm supportive of disabling most of these, but I've found `getActiveCount()` is a helpful tool to monitor load. Thoughts on using a separate property for a subset of these?

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1550,10 +1515,8 @@ private QNode getOrAddNode() {
                         return taskNode;
                     }
                 } else if (headNext instanceof PoolThreadNode || headNext == null) {
-                    if (newNode == null) {
-                        newNode = new PoolThreadNode(Thread.currentThread());
-                    }
                     newNode.setNext(headNext);
```
carterkozak: Can this reference become stale and prevent a chain of nodes from getting garbage collected after threads are deallocated?

dmlloyd: The node might stay around for as long as the thread stays around.  The node *should* be removed from the chain as soon as something is submitted to it, otherwise.  It might be somewhat stale if no tasks are submitted for a while.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
dmlloyd: Yes, they have a pretty substantial performance overhead.  I have a note to see if this method can be implemented by returning the size of the active threads set though...

dmlloyd:
There's still more I want to do here in a second pass, including re-reviewing the other outstanding PRs, but for now I think this is OK to start with.

dmlloyd:
It looks like reusing the thread node instance introduces a race condition that I had forgotten about.  I'll have to back that one out for now.  But it might be possible to come up with some other clever solution in this area.

carterkozak:
That makes sense, I think the improvement to reuse a single PoolThreadNode for each invocation of getOrAddNode head node buys us the majority of the improvement.
I can't thank you enough for your work digging into this!

dmlloyd:
I have one correct solution but it performs worse than just allocating the nodes... trying a new idea now.

dmlloyd:
OK I've pushed up what seems like a good alternative mitigation.  Instead of reusing the instance, we instead avoid allocating the instance by spinning a few times looking for queued tasks.  Creating queued task nodes seems to be cheaper overall compared to creating pooled thread nodes, so I'm seeing numbers very close to the shared instance solution.

carterkozak:
I haven't had a change to run a full benchmark across several core counts, but with all cores available there doesn't appear to be much difference between spinning when headNext is null and allocating the new PoolThreadNode. When I disable yields in getOrAddNode, throughput drops by a factor of five, it appears to perform a bit better with a single yield and no on-spin-wait.

carterkozak:
Perhaps a structure matching `lockTail` to yield after N (full) spins? This appears to perform well in contended cases with few cores as well as high core count scenarios, though I haven't tested exhaustively.

```diff
if (UPDATE_STATISTICS) spinMisses.increment();
- JDKSpecific.onSpinWait();
+ if (spins == YIELD_SPINS) {
+     spins = 0;
+     Thread.yield();
+ } else {
+     spins++;
+     JDKSpecific.onSpinWait();
+ }
```

dmlloyd:
Could you try rerunning your test but adding `prestartAllCoreThreads` at the init section?

carterkozak:
Oh good call, you're absolutely right, it was a benchmarking problem.

carterkozak:
Results (with prestartAllCoreThreads)
```
Benchmark                           (cores)  (executorThreads)             (factory)   Mode  Cnt        Score        Error  Units
ExecutorBenchmarks.benchmarkSubmit        1              16,16  THREAD_POOL_EXECUTOR  thrpt    4   326885.907 ±  48779.926  ops/s
ExecutorBenchmarks.benchmarkSubmit        1              16,16         EQE_SPIN_LOCK  thrpt    4   335704.455 ±   8393.237  ops/s
ExecutorBenchmarks.benchmarkSubmit        1              16,16          EQE_NO_LOCKS  thrpt    4   336362.814 ±  20513.871  ops/s
ExecutorBenchmarks.benchmarkSubmit        1            100,200  THREAD_POOL_EXECUTOR  thrpt    4   282110.556 ±  64286.013  ops/s
ExecutorBenchmarks.benchmarkSubmit        1            100,200         EQE_SPIN_LOCK  thrpt    4   163751.924 ±   6637.838  ops/s
ExecutorBenchmarks.benchmarkSubmit        1            100,200          EQE_NO_LOCKS  thrpt    4   165477.995 ±   6259.012  ops/s
ExecutorBenchmarks.benchmarkSubmit        2              16,16  THREAD_POOL_EXECUTOR  thrpt    4   340278.023 ±  81523.306  ops/s
ExecutorBenchmarks.benchmarkSubmit        2              16,16         EQE_SPIN_LOCK  thrpt    4   494042.241 ±  82964.247  ops/s
ExecutorBenchmarks.benchmarkSubmit        2              16,16          EQE_NO_LOCKS  thrpt    4    68669.688 ±   1910.241  ops/s
ExecutorBenchmarks.benchmarkSubmit        2            100,200  THREAD_POOL_EXECUTOR  thrpt    4   314608.709 ±  32935.210  ops/s
ExecutorBenchmarks.benchmarkSubmit        2            100,200         EQE_SPIN_LOCK  thrpt    4   214875.360 ±   6787.881  ops/s
ExecutorBenchmarks.benchmarkSubmit        2            100,200          EQE_NO_LOCKS  thrpt    4   222698.520 ±   4751.674  ops/s
ExecutorBenchmarks.benchmarkSubmit        4              16,16  THREAD_POOL_EXECUTOR  thrpt    4   397981.410 ±  14230.486  ops/s
ExecutorBenchmarks.benchmarkSubmit        4              16,16         EQE_SPIN_LOCK  thrpt    4   662895.480 ± 206318.781  ops/s
ExecutorBenchmarks.benchmarkSubmit        4              16,16          EQE_NO_LOCKS  thrpt    4   136524.246 ±  35828.220  ops/s
ExecutorBenchmarks.benchmarkSubmit        4            100,200  THREAD_POOL_EXECUTOR  thrpt    4   399977.634 ±   6139.233  ops/s
ExecutorBenchmarks.benchmarkSubmit        4            100,200         EQE_SPIN_LOCK  thrpt    4   373584.557 ±   8727.940  ops/s
ExecutorBenchmarks.benchmarkSubmit        4            100,200          EQE_NO_LOCKS  thrpt    4   396951.744 ±  10269.730  ops/s
ExecutorBenchmarks.benchmarkSubmit        8              16,16  THREAD_POOL_EXECUTOR  thrpt    4   436069.687 ±  19900.733  ops/s
ExecutorBenchmarks.benchmarkSubmit        8              16,16         EQE_SPIN_LOCK  thrpt    4   905448.751 ± 343252.455  ops/s
ExecutorBenchmarks.benchmarkSubmit        8              16,16          EQE_NO_LOCKS  thrpt    4   287289.092 ±  74606.509  ops/s
ExecutorBenchmarks.benchmarkSubmit        8            100,200  THREAD_POOL_EXECUTOR  thrpt    4   429536.213 ±  19178.935  ops/s
ExecutorBenchmarks.benchmarkSubmit        8            100,200         EQE_SPIN_LOCK  thrpt    4   770981.912 ±  55407.537  ops/s
ExecutorBenchmarks.benchmarkSubmit        8            100,200          EQE_NO_LOCKS  thrpt    4   765121.646 ±  50770.996  ops/s
ExecutorBenchmarks.benchmarkSubmit       14              16,16  THREAD_POOL_EXECUTOR  thrpt    4   756642.850 ±  38074.182  ops/s
ExecutorBenchmarks.benchmarkSubmit       14              16,16         EQE_SPIN_LOCK  thrpt    4   745786.134 ±  18988.730  ops/s
ExecutorBenchmarks.benchmarkSubmit       14              16,16          EQE_NO_LOCKS  thrpt    4   486456.341 ±  93098.951  ops/s
ExecutorBenchmarks.benchmarkSubmit       14            100,200  THREAD_POOL_EXECUTOR  thrpt    4   706600.012 ±  62632.633  ops/s
ExecutorBenchmarks.benchmarkSubmit       14            100,200         EQE_SPIN_LOCK  thrpt    4  1355230.918 ±  73210.760  ops/s
ExecutorBenchmarks.benchmarkSubmit       14            100,200          EQE_NO_LOCKS  thrpt    4  1348659.865 ±  73993.616  ops/s
ExecutorBenchmarks.benchmarkSubmit       28              16,16  THREAD_POOL_EXECUTOR  thrpt    4   755386.272 ± 196113.791  ops/s
ExecutorBenchmarks.benchmarkSubmit       28              16,16         EQE_SPIN_LOCK  thrpt    4   977602.904 ±  64685.126  ops/s
ExecutorBenchmarks.benchmarkSubmit       28              16,16          EQE_NO_LOCKS  thrpt    4   562230.945 ±  63242.793  ops/s
ExecutorBenchmarks.benchmarkSubmit       28            100,200  THREAD_POOL_EXECUTOR  thrpt    4   754293.778 ±  27456.244  ops/s
ExecutorBenchmarks.benchmarkSubmit       28            100,200         EQE_SPIN_LOCK  thrpt    4  1664496.164 ± 112363.596  ops/s
ExecutorBenchmarks.benchmarkSubmit       28            100,200          EQE_NO_LOCKS  thrpt    4  1793120.856 ±  54758.939  ops/s
```

carterkozak:
I've opened #70 with an idea to reduce node allocation cost, and in the cases it does occur move the allocation to `waitingForTask` where the thread is otherwise idle.

carterkozak:
Results from that PR on 63a23c2f18ff8ee7ea69c8108803c7f5e582e713

```
Benchmark                           (cores)  (executorThreads)             (factory)   Mode  Cnt        Score        Error  Units
ExecutorBenchmarks.benchmarkSubmit        1              16,16  THREAD_POOL_EXECUTOR  thrpt    4   357682.164 ±  73315.049  ops/s
ExecutorBenchmarks.benchmarkSubmit        1              16,16         EQE_SPIN_LOCK  thrpt    4   365817.909 ±   2724.391  ops/s
ExecutorBenchmarks.benchmarkSubmit        1              16,16          EQE_NO_LOCKS  thrpt    4   361089.122 ±   3582.722  ops/s
ExecutorBenchmarks.benchmarkSubmit        1            100,200  THREAD_POOL_EXECUTOR  thrpt    4   306940.277 ±    309.173  ops/s
ExecutorBenchmarks.benchmarkSubmit        1            100,200         EQE_SPIN_LOCK  thrpt    4   172552.128 ±  42188.384  ops/s
ExecutorBenchmarks.benchmarkSubmit        1            100,200          EQE_NO_LOCKS  thrpt    4   178910.538 ±   9744.082  ops/s
ExecutorBenchmarks.benchmarkSubmit        2              16,16  THREAD_POOL_EXECUTOR  thrpt    4   384507.723 ±  24560.707  ops/s
ExecutorBenchmarks.benchmarkSubmit        2              16,16         EQE_SPIN_LOCK  thrpt    4   537654.439 ± 185820.891  ops/s
ExecutorBenchmarks.benchmarkSubmit        2              16,16          EQE_NO_LOCKS  thrpt    4    77976.078 ±   1310.068  ops/s
ExecutorBenchmarks.benchmarkSubmit        2            100,200  THREAD_POOL_EXECUTOR  thrpt    4   355203.596 ±  11777.332  ops/s
ExecutorBenchmarks.benchmarkSubmit        2            100,200         EQE_SPIN_LOCK  thrpt    4   229392.699 ±   3547.131  ops/s
ExecutorBenchmarks.benchmarkSubmit        2            100,200          EQE_NO_LOCKS  thrpt    4   235271.868 ±   6001.266  ops/s
ExecutorBenchmarks.benchmarkSubmit        4              16,16  THREAD_POOL_EXECUTOR  thrpt    4   443271.467 ±  23246.743  ops/s
ExecutorBenchmarks.benchmarkSubmit        4              16,16         EQE_SPIN_LOCK  thrpt    4   662697.561 ± 224435.239  ops/s
ExecutorBenchmarks.benchmarkSubmit        4              16,16          EQE_NO_LOCKS  thrpt    4   165462.929 ±  16453.597  ops/s
ExecutorBenchmarks.benchmarkSubmit        4            100,200  THREAD_POOL_EXECUTOR  thrpt    4   440952.600 ±  21476.754  ops/s
ExecutorBenchmarks.benchmarkSubmit        4            100,200         EQE_SPIN_LOCK  thrpt    4   411013.918 ±  21794.301  ops/s
ExecutorBenchmarks.benchmarkSubmit        4            100,200          EQE_NO_LOCKS  thrpt    4   429880.983 ±  26201.831  ops/s
ExecutorBenchmarks.benchmarkSubmit        8              16,16  THREAD_POOL_EXECUTOR  thrpt    4   474214.980 ±  18200.269  ops/s
ExecutorBenchmarks.benchmarkSubmit        8              16,16         EQE_SPIN_LOCK  thrpt    4   937372.156 ± 392849.675  ops/s
ExecutorBenchmarks.benchmarkSubmit        8              16,16          EQE_NO_LOCKS  thrpt    4   352997.239 ±  22343.079  ops/s
ExecutorBenchmarks.benchmarkSubmit        8            100,200  THREAD_POOL_EXECUTOR  thrpt    4   467673.690 ±   8560.542  ops/s
ExecutorBenchmarks.benchmarkSubmit        8            100,200         EQE_SPIN_LOCK  thrpt    4   602931.815 ±  24965.591  ops/s
ExecutorBenchmarks.benchmarkSubmit        8            100,200          EQE_NO_LOCKS  thrpt    4   877563.840 ±  62904.107  ops/s
ExecutorBenchmarks.benchmarkSubmit       14              16,16  THREAD_POOL_EXECUTOR  thrpt    4   779463.583 ±  46802.117  ops/s
ExecutorBenchmarks.benchmarkSubmit       14              16,16         EQE_SPIN_LOCK  thrpt    4  1169717.353 ±  64192.565  ops/s
ExecutorBenchmarks.benchmarkSubmit       14              16,16          EQE_NO_LOCKS  thrpt    4   617351.086 ±  69232.530  ops/s
ExecutorBenchmarks.benchmarkSubmit       14            100,200  THREAD_POOL_EXECUTOR  thrpt    4   756646.686 ±  33296.965  ops/s
ExecutorBenchmarks.benchmarkSubmit       14            100,200         EQE_SPIN_LOCK  thrpt    4  1496606.372 ±  88772.949  ops/s
ExecutorBenchmarks.benchmarkSubmit       14            100,200          EQE_NO_LOCKS  thrpt    4  1498699.083 ± 159938.396  ops/s
ExecutorBenchmarks.benchmarkSubmit       28              16,16  THREAD_POOL_EXECUTOR  thrpt    4   754328.339 ± 305957.771  ops/s
ExecutorBenchmarks.benchmarkSubmit       28              16,16         EQE_SPIN_LOCK  thrpt    4  1645558.945 ±  34785.244  ops/s
ExecutorBenchmarks.benchmarkSubmit       28              16,16          EQE_NO_LOCKS  thrpt    4   726105.034 ± 119425.935  ops/s
ExecutorBenchmarks.benchmarkSubmit       28            100,200  THREAD_POOL_EXECUTOR  thrpt    4   741269.216 ±  16412.150  ops/s
ExecutorBenchmarks.benchmarkSubmit       28            100,200         EQE_SPIN_LOCK  thrpt    4  1710123.573 ±  17327.327  ops/s
ExecutorBenchmarks.benchmarkSubmit       28            100,200          EQE_NO_LOCKS  thrpt    4  2020520.405 ±  43268.846  ops/s
```

PR #65, opened 2020-03-16
JBTHR-83: unpark is never invoked under a lock

carterkozak:
Unpark is an expensive operation, taking many times locker than
other operations run with the exclusive tail lock. This change
moves unpark outside of the locked region to allow work to
continue.

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -762,18 +762,7 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
     public void execute(Runnable runnable) {
         Assert.checkNotNullParam("runnable", runnable);
         final Runnable realRunnable = JBossExecutors.classLoaderPreservingTaskUnchecked(runnable);
-        int result;
-        if (TAIL_LOCK) {
-            Lock lock = this.tailLock;
-            lock.lock();
-            try {
-                result = tryExecute(realRunnable);
-            } finally {
-                lock.unlock();
-            }
-        } else {
-            result = tryExecute(realRunnable);
-        }
+        int result = tryExecute(realRunnable);
         boolean ok = false;
         if (result == EXE_OK) {
             // last check to ensure that there is at least one existent thread to avoid rare thread timeout race condition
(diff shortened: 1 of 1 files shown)


carterkozak:
Github isn't rendering the diff very nicely, I'll annotate inline

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -762,18 +762,7 @@ public Builder setMBeanName(final String mBeanName) {
...
-            Lock lock = this.tailLock;
-            lock.lock();
-            try {
-                result = tryExecute(realRunnable);
-            } finally {
-                lock.unlock();
-            }
-        } else {
-            result = tryExecute(realRunnable);
-        }
```
carterkozak: locks moved into tryExecute

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1727,79 +1716,70 @@ boolean doStartThread(Runnable runnable) throws RejectedExecutionException {
 
     private int tryExecute(final Runnable runnable) {
         QNode tailNext;
+        PoolThreadNode toUnpark = null;
```
carterkozak: This allows us to effectively return both the result `int` and a `PoolThreadNode` to be unparked outside of the lock.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1727,79 +1716,70 @@ boolean doStartThread(Runnable runnable) throws RejectedExecutionException {
...
-                // post-actions (fail):
-                //   retry with new tail(snapshot)
-                if (tail.compareAndSetNext(tailNext, tailNextNext)) {
-                    PoolThreadNode consumerNode = (PoolThreadNode) tailNext;
-                    // state change ex2:
-                    //   tail(snapshot).next(snapshot).task ← runnable
-                    // succeeds: ex1
+        ExtendedLock tailLock = this.tailLock;
+        if (TAIL_LOCK) tailLock.lock();
+        try {
```
carterkozak: 3 new lines, below is tabbed for the try block

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1727,79 +1716,70 @@ boolean doStartThread(Runnable runnable) throws RejectedExecutionException {
...
+                        // succeeds: ex1
+                        // preconditions:
+                        //   tail(snapshot).next(snapshot).task = WAITING
+                        // post-actions (succeed):
+                        //   unpark thread and return
+                        // post-actions (fail):
+                        //   retry outer with new tail(snapshot)
+                        if (consumerNode.compareAndSetTask(WAITING, runnable)) {
+                            // unpark must be executed without a lock
+                            toUnpark = consumerNode;
```
carterkozak:
```diff
- consumerNode.unpark();
+ // unpark must be executed without a lock
+ toUnpark = consumerNode;
```

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1727,79 +1716,70 @@ boolean doStartThread(Runnable runnable) throws RejectedExecutionException {
...
-                }
-                assert tr == AT_NO;
-                // no; try to enqueue
-                if (! NO_QUEUE_LIMIT && ! increaseQueueSize()) {
-                    // queue is full
-                    // OK last effort to create a thread, disregarding growth limit
-                    tr = tryAllocateThread(0.0f);
+                    // retry with new tail(snapshot) as was foretold
+                    tail = this.tail;
+                    if (UPDATE_STATISTICS) spinMisses.increment();
```
carterkozak: hmm, while we're here it probably makes sense to increment stats prior to updating tail for a tight spin.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1809,43 +1789,61 @@ private int tryExecute(final Runnable runnable) {
...
-                // no consumers are waiting and the tail(snapshot).next node is non-null and not a task node, therefore it must be a...
-                assert tailNext instanceof TerminateWaiterNode;
-                // shutting down
-                result = EXE_REJECT_SHUTDOWN;
-                break;
             }
+        } finally {
+            if (TAIL_LOCK) tailLock.unlock();
         }
+        if (toUnpark != null) toUnpark.unpark();
```
carterkozak: finally + unpark are new.

carterkozak:
Resolved by #62

PR #66, opened 2020-03-16
Update the optimistic park spin to use a deterministic approach

carterkozak:
This algorithm first spins, escalates to yield, then parks after
a fixed number of iterations. This change updates the edge conditions
to avoid cases where we yield without attempting cas operation
afterward.


carterkozak:
Opened as a draft, I still need to do some performance analysis on this change when #65 is present.

carterkozak:
Ready for review, in my testing this outperforms the original implementation by a couple percent. I imagine the probabilistic yield exists because onSpinWait was not available when LinkedTransferQueue was implemented.

carterkozak:
Resolved by #62

PR #69, opened 2020-03-20
JBTHR-84: Statistics methods return -1 when statistics are disabled

carterkozak:


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -630,6 +630,7 @@ public int getMaximumQueueSize() {
 
         /**
          * Set the maximum queue size.
+         * This has no impact when <pre>jboss.threads.eqe.unlimited-queue</pre> is set.
```
dmlloyd: This formats strangely.
```suggestion
         * This has no impact when {@code jboss.threads.eqe.unlimited-queue} is set.
```

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1308,52 +1309,56 @@ public int getQueueSize() {
...
 
     /**
      * Get an estimate of the number of threads which are currently doing work on behalf of the thread pool.
      *
-     * @return the active count estimate
+     * @return the active count estimate or -1 when {@code jboss.threads.eqe.statistics.active-count} is disabled
      */
     public int getActiveCount() {
-        return activeCount;
+        return UPDATE_ACTIVE_COUNT ? activeCount : -1;
```
carterkozak: Updated this to reference `statistics.active-count` as well

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
carterkozak: Good call, thanks!

PR #70, opened 2020-03-21
Eagerly create and reuse PoolThreadNode instances

carterkozak:
When getOrAddNode does not submit a PoolThreadNode it's not necessary
to create more instances.

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -1422,10 +1422,12 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
             // run the initial task
             doRunTask(getAndClearInitialTask());
 
+            // Eagerly allocate a PoolThreadNode for the next time it's needed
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1422,17 +1422,22 @@ public void run() {
...
-                node = getOrAddNode();
+                node = getOrAddNode(nextPoolThreadNode);
                 if (node instanceof TaskNode) {
                     // task node was removed
                     doRunTask(((TaskNode) node).getAndClearTask());
                     continue;
                 } else if (node instanceof PoolThreadNode) {
                     // pool thread node was added
                     final PoolThreadNode newNode = (PoolThreadNode) node;
+                    assert newNode == nextPoolThreadNode;
```
carterkozak: This equality check should replace the instanceof PoolThreadNode check

dmlloyd:
Did this make a measurable difference?

carterkozak:
Oh yes, especially using a thread pool which sees both queueing and waiting threads (my core=16,max=16 case with 32 threads submitting work). I added concrete benchmarks in #62

dmlloyd:
Ah, great!

PR #72, opened 2020-04-18
tryExecute retry mode under contention if !TAIL_LOCK

franz1981:
I've noticed that the !TAIL_LOCK case can see a huge amount of time spent awaiting tail::getNext instanceof TaskNode == false. 
When there is a big amount of contending producers there is no benefit to chase tail::getNext in a (nearly) tight loop, given that the `instance of` check would likely cause anyway a cache miss to read the class world: similarly to  the changes to "relax" the contention while spin locking, this PR is trying to not break anything and use a similar strategy to guarantee some progress instead of consuming unnecessary CPU in a (nearly) tight loop.

PR #73, opened 2020-04-21
Optimize the tryExecute lock-free queue handling

carterkozak:
This is based on a discussion from
https://github.com/jbossas/jboss-threads/pull/72


franz1981:
I believe this approach was used in the earliest version of `ConcurrentLinkedQueue` by Brian Goetz, see https://github.com/eagle518/jdk-source-code/blob/91b771140de051fb843af246ab826dd6ff688fe3/jdk5.0_src/j2se/src/share/classes/java/util/concurrent/ConcurrentLinkedQueue.java#L179-L196 
```java
    public boolean offer(E o) {
        if (o == null) throw new NullPointerException();
        Node<E> n = new Node<E>(o, null);
        for(;;) {
            Node<E> t = tail;
            Node<E> s = t.getNext();
            if (t == tail) {
                if (s == null) {
                    if (t.casNext(s, n)) {
                        casTail(t, n);
                        return true;
                    }
                } else {
                    casTail(t, s);
                }
            }
        }
    }
```
This same approach has changed and now (jdk 14) the opportunistic update of tail doesn't happen anymore AND we now use `tail` differently, because it's updated every 2 hops (or more), see https://github.com/AdoptOpenJDK/openjdk-jdk/blob/634c676e6c145c936b460debaa95d0b6cfa61ae2/src/java.base/share/classes/java/util/concurrent/ConcurrentLinkedQueue.java#L354-L381,
```java
    public boolean offer(E e) {
        final Node<E> newNode = new Node<E>(Objects.requireNonNull(e));

        for (Node<E> t = tail, p = t;;) {
            Node<E> q = p.next;
            if (q == null) {
                // p is last node
                if (NEXT.compareAndSet(p, null, newNode)) {
                    // Successful CAS is the linearization point
                    // for e to become an element of this queue,
                    // and for newNode to become "live".
                    if (p != t) // hop two nodes at a time; failure is OK
                        TAIL.weakCompareAndSet(this, t, newNode);
                    return true;
                }
                // Lost CAS race to another thread; re-read next
            }
            else if (p == q)
                // We have fallen off list.  If tail is unchanged, it
                // will also be off-list, in which case we need to
                // jump to head, from which all live nodes are always
                // reachable.  Else the new tail is a better bet.
                p = (t != (t = tail)) ? t : head;
            else
                // Check for tail updates after two hops.
                p = (p != t && t != (t = tail)) ? t : q;
        }
    }
```
So, it seems that another approach that reduce the number of `cas` has been chosen (no longer need opportunistic updates) instead and probably `tail` is no longer required to move forward, given that can be reached by `head`, see
```     
     * Both head and tail are permitted to lag.  In fact, failing to
     * update them every time one could is a significant optimization
     * (fewer CASes). As with LinkedTransferQueue (see the internal
     * documentation for that class), we use a slack threshold of two;
     * that is, we update head/tail when the current pointer appears
     * to be two or more steps away from the first/last node.
     *
     * Since head and tail are updated concurrently and independently,
     * it is possible for tail to lag behind head (why not)?
```
In short: your version seems to match what the original version of Brian Goetz was doing (that's ok), but probably with scalability in mind, getting a lighter `offer` by reducing the amount of CAS is a bigger win with > 2 cores eg `compareAndSetTail` can became somthing using `Unsafe::putOrdered` instead, if the offer is the only allowed to move the `tail` after a successfull `compareAndSetNext`.

carterkozak:
Benchmarking results on this change (with UPDATE_TAIL=true, the new default in this PR) https://gist.githubusercontent.com/carterkozak/2800ccb3b57749d5a2666afe6a380664/raw/43d78ae6a37dab7bcb397abfc8968c7bb434e227/full_benchmarks_jbthr73.log

I think it makes sense to remove the UPDATE_TAIL flag entirely and always update the value, otherwise it's possible to lag far behind where we expect (when locks are disabled anyhow).

Great find on the ConcurrentLinkedQueue optimization, I'm impressed! I'll read through it after I've had a proper cup of coffee!

dmlloyd:
This approach seems like it makes sense - rereading tail should be more efficient than chasing up the list - and unless my eyes are crossing, the benchmark results look pretty good as well.  Is this ready @carterkozak or is there anything else you want to examine or test?

FWIW I think it's probably safe to drop spin locks from future consideration.

franz1981:
I dont understand the case with !UPDATE_TAIL; it walk the list till "the end" without doing anything? I prefer it to read `this.tail` instead and retry, because we've already seen that there is no benefit just 
walking the list.

carterkozak:
I'm going to take one more pass over this, I think I can make the inner loop cleaner.
I agree that `UPDATE_TAIL` doesn't make sense to disable anymore, would it be alright if I remove the property entirely?

dmlloyd:
> I'm going to take one more pass over this, I think I can make the inner loop cleaner.
> I agree that `UPDATE_TAIL` doesn't make sense to disable anymore, would it be alright if I remove the property entirely?

Sure, we have the corresponding data now.

franz1981:
@dmlloyd @carterkozak Let me know if I should close https://github.com/jbossas/jboss-threads/pull/72 in favour of this one :+1:

carterkozak:
I'm sure there are additional optimizations that could be beneficial, but this change is meant to minimally fix the problems that resulted in poor performance under contention without changing the algorithm. This should result in a better baseline to compare those future optimizations against :-)

Thanks for the help and feedback @franz1981 and @dmlloyd!

PR #74, opened 2020-04-21
Save the linked list of the executor to suffer from GC nepotism

franz1981:
This has turned to be a relevant GC issue that can cause several problems (including OOM) on  many linked-lists implementations. 
For reference:
- http://psy-lob-saw.blogspot.com/2016/03/gc-nepotism-and-linked-queues.html
- https://github.com/akka/akka/issues/19216#issuecomment-194835210
- https://github.com/AdoptOpenJDK/openjdk-jdk14/blob/f7165c322a6abefa34afa9eeb76dbd56f70aae09/src/java.base/share/classes/java/util/LinkedList.java#L453

I've solved the problem in the same way CLQ explain here: https://github.com/AdoptOpenJDK/openjdk-jdk14/blob/f7165c322a6abefa34afa9eeb76dbd56f70aae09/src/java.base/share/classes/java/util/concurrent/ConcurrentLinkedQueue.java#L149

```
We use the trick of
linking a Node that has just been dequeued to itself.  Such a
self-link implicitly means to advance to head.
```

In addition, I see that this comment gives an additional hint on how to handle producer-side the case where `tail.next == next` ie `self-link implicitly means to advance to head`: 
- if tail is unchanged (from the last read) it means that head is the most up-to-date node to append to and we can use it
- if tail is changed (from the last read) it just means that we've fallen behind and can use the new tail again

I haven't implemented this strategy, but I let it re-attempt reading `this.tail` if a consumer past over a producer.

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -794,9 +794,16 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
         QNode headNext;
         for (;;) {
             headNext = head.getNext();
+            if (headNext == head) {
+                // a racing consumer has already consumed it (and moved head)
+                head = this.head;
+                continue;
+            }
             if (headNext instanceof TaskNode) {
                 TaskNode taskNode = (TaskNode) headNext;
                 if (compareAndSetHead(head, taskNode)) {
+                    // save from GC nepotism
+                    head.setNextOrdered(head);
                     if (! NO_QUEUE_LIMIT) decreaseQueueSize();
                     head = taskNode;
                     list.add(taskNode.task);
@@ -1515,9 +1522,21 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
             for (;;) {
                 head = EnhancedQueueExecutor.this.head;
                 headNext = head.getNext();
+                // this happen if another consumer has already consumed head:
+                // retry with a fresh head
+                if (headNext == head) {
+                    if (UPDATE_STATISTICS) spinMisses.increment();
+                    JDKSpecific.onSpinWait();
+                    continue;
+                }
                 if (headNext instanceof TaskNode) {
                     TaskNode taskNode = (TaskNode) headNext;
                     if (compareAndSetHead(head, taskNode)) {
+                        // save from GC Nepotism: generational GCs don't like
+                        // cross-generational references, so better to "clean-up" head::next
+                        // to save dragging head::next into the old generation.
+                        // Clean-up cannot just null out next
+                        head.setNextOrdered(head);
                         if (! NO_QUEUE_LIMIT) decreaseQueueSize();
                         return taskNode;
                     }
@@ -1525,6 +1544,11 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
                     nextPoolThreadNode.setNextRelaxed(headNext);
                     if (head.compareAndSetNext(headNext, nextPoolThreadNode)) {
                         return nextPoolThreadNode;
+                    } else if (headNext != null) {
+                        // GC Nepotism:
+                        // save dragging headNext into old generation
+                        // (although being a PoolThreadNode it won't make a big difference)
+                        nextPoolThreadNode.setNextRelaxed(null);
                     }
                 } else {
                     assert headNext instanceof TerminateWaiterNode;
@@ -1701,6 +1725,15 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
         TaskNode node = null;
         for (;;) {
             tailNext = tail.getNext();
+            // tail is already consumed
+            if (tailNext == tail) {
+                if (UPDATE_STATISTICS) spinMisses.increment();
+                JDKSpecific.onSpinWait();
+                // retry with new tail(snapshot)
+                tail = this.tail;
+                continue;
+            }
+            assert tailNext != tail;
             if (tailNext instanceof TaskNode) {
                 TaskNode tailNextTaskNode = (TaskNode) tailNext;
                 // Opportunistically update tail to the next node. If this operation has been handled by
@@ -1745,6 +1778,10 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
                     // post-actions (fail):
                     //   retry outer with new tail(snapshot)
                     if (consumerNode.compareAndSetTask(WAITING, runnable)) {
+                        // GC Nepotism:
+                        // We can save consumerNode::next from being dragged into
+                        // old generation, if possible
+                        consumerNode.compareAndSetNext(tailNextNext, null);
                         consumerNode.unpark();
                         return EXE_OK;
                     }
@@ -2088,6 +2125,10 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
 
         void setNextRelaxed(final QNode node) {
             unsafe.putObject(this, nextOffset, node);
+        }
+
+        void setNextOrdered(final QNode node) {
+            unsafe.putOrderedObject(this, nextOffset, node);
         }
     }
 


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1706,17 +1714,30 @@ private int tryExecute(final Runnable runnable) {
...
-                do {
-                    if (UPDATE_STATISTICS) spinMisses.increment();
-                    tailNextTaskNode = (TaskNode) tailNext;
-                    // retry
-                    tail = tailNextTaskNode;
-                    tailNext = tail.getNext();
-                } while (tailNext instanceof TaskNode);
-                // opportunistically update for the possible benefit of other threads
-                if (UPDATE_TAIL) compareAndSetTail(tail, tailNextTaskNode);
+                if (UPDATE_TAIL) {
```
franz1981, line 1727: @carterkozak I've used the same approach of you on https://github.com/jbossas/jboss-threads/pull/73#issuecomment-616953067, but saving to walk the list if UPDATE_TAIL won't validate (using the compareAndSet) that we're lagging behind `this.tail`.
I've changed this code path as well because the walking algorithm would have changed due to the chance to get  a recursive next pointer when a node is consumed.

franz1981:
@dmlloyd @carterkozak 
I was sure to have broken (maybe I am anyway eh) the current logic by this change, but seems that I'm able to make fail to master with this test (on my 4 cores with HT laptop):
```java
@Test
    public void gcNepotism() throws ExecutionException, InterruptedException {
        System.setProperty("jboss.threads.eqe.park-spins", Long.toString(Integer.MAX_VALUE));
        System.setProperty("jboss.threads.eqe.tail-lock", Boolean.FALSE.toString());
        Assume.assumeFalse(EnhancedQueueExecutor.TAIL_LOCK);
        EnhancedQueueExecutor executor = (new EnhancedQueueExecutor.Builder())
                .setCorePoolSize(10)
                .setMaximumPoolSize(20)
                .build();
        final int producers = 10;
        int total = 100_000;
        ExecutorService service = Executors.newFixedThreadPool(producers);
        Future<?>[] tasks = new Future[producers];
        for (int i = 0; i<producers;i++) {
            tasks[i] = service.submit(() -> {
                for (int j = 0; j < total; j++) {
                    executor.execute(() -> {
                    });
                }
            });
        }
        for (int i = 0; i<producers;i++) {
            tasks[i].get();
        }
        executor.submit(()->{}).get();
        service.shutdown();
        executor.shutdown();
    }
```
On `mater` I'm getting:
```java
java.util.concurrent.ExecutionException: java.lang.AssertionError

	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
	at org.jboss.threads.EnhancedQueueExecutorTest.gcNepotism(EnhancedQueueExecutorTest.java:179)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:230)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:58)
Caused by: java.lang.AssertionError
	at org.jboss.threads.EnhancedQueueExecutor.tryExecute(EnhancedQueueExecutor.java:1740)
	at org.jboss.threads.EnhancedQueueExecutor.execute(EnhancedQueueExecutor.java:755)
	at org.jboss.threads.EnhancedQueueExecutorTest.lambda$gcNepotism$1(EnhancedQueueExecutorTest.java:173)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:266)
	at java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```
ie it's failing on `assert tail instanceof TaskNode && tail.task == null;` on the `tail.task == null` check

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1749,6 +1778,10 @@ private int tryExecute(final Runnable runnable) {
                     // post-actions (fail):
                     //   retry outer with new tail(snapshot)
                     if (consumerNode.compareAndSetTask(WAITING, runnable)) {
+                        // GC Nepotism:
+                        // We can save consumerNode::next from being dragged into
+                        // old generation, if possible
+                        consumerNode.compareAndSetNext(tailNextNext, null);
```
franz1981, line 1784: I've used a compareAndSetNext because I'm not sure that this case is single-writer, maybe yes.

franz1981:
If I need to provide a test to make evident the GC nepotism problem I can do it as well :+1

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
carterkozak: Unfortunately I think we have to walk the list to avoid blocking on execute. I don't think UPDATE_TAIL should ever be disabled, see https://github.com/jbossas/jboss-threads/pull/72#discussion_r412149895

franz1981: I'm answering directly there :)

franz1981: @dmlloyd @carterkozak Regardless the way I've written this specific code path: the GC and the assertion issue on https://github.com/jbossas/jboss-threads/pull/74#issuecomment-617026360 are still valid; Probably will rebase this code path on what will be merged after  https://github.com/jbossas/jboss-threads/pull/73

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -794,9 +794,11 @@ public void shutdown() {
         QNode headNext;
         for (;;) {
             headNext = head.getNext();
-            if (headNext instanceof TaskNode) {
+            if (headNext != head && headNext instanceof TaskNode) {
                 TaskNode taskNode = (TaskNode) headNext;
                 if (compareAndSetHead(head, taskNode)) {
+                    // save from GC nepotism
+                    head.setNextRelaxed(head);
```
dmlloyd, line 801: Can you prove that this can never cause nodes to be skipped?

franz1981: good point: looking at the original code I see that is draining `TaskNode`s until `tail::getNext` is null, after a shutdown: I can preserve the original logic to handle the case of a concurrent call to it.

franz1981: I've changed the visibility guarantess while setting next too, taking "inspiration" from CLQ:
https://github.com/openjdk-mirror/jdk7u-jdk/blob/8aaca72128077594ecf4ecdee761e78d4ecb8b7d/src/share/classes/java/util/concurrent/ConcurrentLinkedQueue.java#L305

```java
    final void updateHead(Node<E> h, Node<E> p) {
        if (h != p && casHead(h, p))
            h.lazySetNext(h);
    }
```

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1515,16 +1522,33 @@ private QNode getOrAddNode(PoolThreadNode nextPoolThreadNode) {
...
+                    continue;
+                }
                 if (headNext instanceof TaskNode) {
                     TaskNode taskNode = (TaskNode) headNext;
                     if (compareAndSetHead(head, taskNode)) {
+                        // save from GC Nepotism: generational GCs don't like
+                        // cross-generational references, so better to "clean-up" head::next
+                        // to save dragging head::next into the old generation.
+                        // Clean-up cannot just null out next
+                        head.setNextOrdered(head);
```
dmlloyd, line 1539: I need to see a correctness proof here.  What are the possible observable output states, and why?  Are the states all valid and already described above?  What are the preconditions and postconditions?

franz1981: It's mostly expressed in term of acquire-release relationships between atomic operations: 

- only one thread can win compareAndSetHead(head, taskNode) for a specific head, taskNode pair
- other threads can reach head while its head::next is valued differently:

1. "slow" producers (ie producers descheduled after reading this.tail for enough time that other producers have moved on and a consumer has already moved head): 
- before `head.setNextOrdered(head)`: they will try to `compareAndSetNext(null, newValue)` and will fail, because given that the consumer has been able to move past to it, it means that next != null
- after `head.setNextOrdered(head)`: they will ry to `compareAndSetNext(null, newValue)` and will fail, because head != null
2. slow consumers (ie consumers that have been descheduled after reading this.head and are going to read head::getNext)
- before `head.setNextOrdered(head)`: they will get the same value of the consumer that has won casHead, hence they will attempts casHead, failing because has been already moved on
- after `head.setNextOrdered(head)`: they will get the new value (head::next == head) and they will reattempt reading this.head. Given that head.setNextOrdered is write releasing any previous write (including the won casHead), load head::next and finding == head is load acquiring the new value of head, hence reading head will return *at least* the value released by the consumer that has won casHead

franz1981: The states already written in the overview of the class should be valid, because I've tried hard to not break the way we use `null` and just add checks that head::next == next means that "something" has changed: in particular `head` and by consequence `tail`.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1515,16 +1522,33 @@ private QNode getOrAddNode(PoolThreadNode nextPoolThreadNode) {
...
                     }
                 } else if (headNext instanceof PoolThreadNode || headNext == null) {
                     nextPoolThreadNode.setNextRelaxed(headNext);
                     if (head.compareAndSetNext(headNext, nextPoolThreadNode)) {
                         return nextPoolThreadNode;
+                    } else if (headNext != null) {
+                        // GC Nepotism:
+                        // save dragging headNext into old generation
+                        // (although being a PoolThreadNode it won't make a big difference)
+                        nextPoolThreadNode.setNextRelaxed(null);
```
dmlloyd, line 1551: Am I correct in understanding that this is presumed safe because `nextPoolThreadNode` has not yet been published?

franz1981: Excactly and because `nextPoolThreadNode` (if not published) is still owned by a single thread (the consumer thread itself), hence `setNextRelaxed` is fine,because no other threads would never be able to read `nextPoolThreadNode::next` concurrently.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
dmlloyd: OK I think I understand all the possibilities here. But couldn't this be `setNextRelaxed`? We're not using `node.next == node` as the solitary indicator that it has been removed, so it's probably OK if some threads don't observe that (they would just spin past the node using the current behavior).  Does it make a difference performance-wise?

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
dmlloyd: Maybe we should do a weakCompareAndSet here to avoid the heavier memory barrier?

franz1981: IMO is safe wekCompareAndSet, but probably will make any difference only for ARM/PowerPC servers: AFAIK newish x86s won't have any specific code path implemented on the OpenJDK for this (but I could be wrong)

dmlloyd: OK.  Well we do care about ARM and AArch64 so I think that makes it valid.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
franz1981: > Does it make a difference performance-wise?
setNextOrdered on x86 isn't using any specific hardware barrier; thanks to the strong memory model of x86 and the only perf hit would have been by using "updaters" but yet negligible.

From a correctness pov yes: the only thing we care about here is *atomicity*, but sadly `Unsafe` isn't providing the right methods to just have that: ideally this is what we need here https://docs.oracle.com/javase/9/docs/api/java/lang/invoke/VarHandle.AccessMode.html#SET_OPAQUE , because the compareAndSet previous to it already provide the right sequential consistent write we need.

dmlloyd: Ah, I see.

franz1981:
@dmlloyd I'm concerned about the asssert on https://github.com/jbossas/jboss-threads/pull/74#issuecomment-617026360 that's failing on `master` pre-lock removal too. Is an asssertion still valid?

dmlloyd:
> ie it's failing on `assert tail instanceof TaskNode && tail.task == null;` on the `tail.task == null` check
> That assertion is still valid?

The assertion is based on `tail` being a dead task node.  I'd say the assertion is still valid.  If `tail` is not a dead task, how does it come to pass that the task was not executed by the current thread?

franz1981:
> > ie it's failing on `assert tail instanceof TaskNode && tail.task == null;` on the `tail.task == null` check
> > That assertion is still valid?
> 
> The assertion is based on `tail` being a dead task node. I'd say the assertion is still valid. If `tail` is not a dead task, how does it come to pass that the task was not executed by the current thread?

Agree: in this case, the test I've written was failing on `master` pre TAIL_LOCK removal, is failing on `master` now and is failing on this PR as well...

carterkozak:
This may be naive, but isn't it possible that tail is destined to become a dead task node but getAndClearTask hasn't run yet? In this case I think head may have progressed past the tail, but should self-correct.

Aside from this assert I'm curious why TaskNode.task is volatile.

dmlloyd:
IIRC it was originally volatile because that was the means of publishing the task from the producer to the consumer.  But maybe it no longer needs to be; if someone has time to do the requisite analysis I'd be open to the change.

carterkozak:
Running the `gcNepotism` test on 3.1.0.Final I'm getting assertion errors with and without locks (modified to allow the test to run with tail-lock enabled), with and without park-spins. The failure rate isn't quite as high as without locks, but every two to three runs.

carterkozak:
I've started a round of benchmarks, I'll post the results here in the morning.

franz1981:
@carterkozac Re the assertion error: so do you think it worths to investigate and fix it, although is there from some time?

carterkozak:
Full suite of benchmarks for raw comparison (still using UseParallelOldGC):

https://gist.github.com/carterkozak/0f265ca33bd668b4f1641da362f613dd

Using `-XX:+UseG1GC -XX:-UseCondCardMark`:
```
Benchmark                           (burstSize)  (cores)  (executorThreads)     (factory)  (spinWaitCompletion)  Mode  Cnt    Score   Error  Units
ExecutorBenchmarks.benchmarkSubmit           10       28              16,16  EQE_NO_LOCKS                 false  avgt    4  177.902 ± 3.426  us/op
```
Master:
```
Benchmark                           (burstSize)  (cores)  (executorThreads)     (factory)  (spinWaitCompletion)  Mode  Cnt    Score   Error  Units
ExecutorBenchmarks.benchmarkSubmit           10       28              16,16  EQE_NO_LOCKS                 false  avgt    4  175.534 ± 5.646  us/op
```

Using `-XX:+UseG1GC -XX:+UseCondCardMark`:

this PR:
```
Benchmark                           (burstSize)  (cores)  (executorThreads)     (factory)  (spinWaitCompletion)  Mode  Cnt    Score    Error  Units
ExecutorBenchmarks.benchmarkSubmit           10       28              16,16  EQE_NO_LOCKS                 false  avgt    4  178.436 ± 10.158  us/op
```
Master:
```
Benchmark                           (burstSize)  (cores)  (executorThreads)     (factory)  (spinWaitCompletion)  Mode  Cnt    Score    Error  Units
ExecutorBenchmarks.benchmarkSubmit           10       28              16,16  EQE_NO_LOCKS                 false  avgt    4  176.804 ± 13.321  us/op
```

There's not much difference in the values, but our tasks run very quickly and nodes die young for the most part.

franz1981:
@carterkozak  Parallel GC needs UseCondCardMark to perform well on high-core machines. G1GC already have conditional card marks by default: regardless tasks getting old in human metrics, it mostly depends on the (young) heap size then the actional lifecycle how is the promotion rate. Probably worth using GCProfiler or (-prof gc) on the bench IMO

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
carterkozak: Is this safe in cases where `EQE.head` passes `EQE.tail`?

The tail reference may lag (currently by only one node, but the potential jdk14 CQL optimization could potentially allow more lag), which I think allows `EQE.head` to be `EQE.tail.next`. Clearing references may unlink tail from head and orphan submitted tasks.

It's very possible that I've missed an important piece here, I'm not confident that I've gotten this right.

dmlloyd:
Just want to note that the assertion problem does require a solution.  I'm looking into it, but you are all welcome to look as well (chances are it will take me longer to find the problem anyway).

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
franz1981: It shouldn't AFAIK.
A producer set prev_tail::casNext(null, new_tail) and then casTail(prev_tail,new_tail), while a consumer can consume only prev_tail, not new_tail (yet).
Given that other producers are able to read `this.tail` (ie new_tail) and evetually `this.tail == new_tail` others are always able to append new element without loosing bits around

dmlloyd:
Well, I solved the assert mystery.  It is indeed invalid (see #75 for the explanation).

dmlloyd:
Any other remarks before I merge?

carterkozak:
Looks great, thanks!

I've opened a PR to address TaskNode.task volatility #76 describing why I believe it's a safe change.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
franz1981: @carterkozak Thinking about it again and again, probably we're "lucky" that we cannot lag, but if we would start to use the jdk 14 optimization (as you've suggested) we need, on 
 producer side, chase the last tail starting from head (that in theory should be near(ish) given that has past tail, that has lagged behind, if node.next == node (see https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/concurrent/ConcurrentLinkedQueue.java#L348).

nitsanw: > From a correctness pov yes: the only thing we care about here is atomicity
@franz1981 you can use a plain write in this case, since reference writes are always atomic

franz1981: @nitsanw thanks, yep JLS should ensure that things ref equality should work fine here, no tearing on the reference value indeed, thanks!

PR #75, opened 2020-04-23
Fix faulty assertion

dmlloyd:
The assertion is invalid because if there are two worker threads, and one worker has updated the head pointer to a populated task node, while in the meantime the second thread observes the end of the list and adds a pooled thread node to the end, then tryExecute may observe a 'dead' task node which nevertheless still contains a task.


carterkozak:
Thank you!

PR #79, opened 2020-04-29
Restore previous class loader after safeRun

dmlloyd:


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -2047,7 +2048,7 @@ void safeRun(final Runnable task) {
                 // nothing else we can safely do here
             }
         } finally {
-            JBossExecutors.clearContextClassLoader(currentThread);
+            JBossExecutors.setContextClassLoader(currentThread, oldCl);
```
carterkozak, line 2051: I thought ContextClassLoaderSavingRunnable already took care of this, do we need both `JBossExecutors.classLoaderPreservingTaskUnchecked(runnable);` in execute and the custom handling here?

dmlloyd: `safeRun` also runs during termination and could potentially be called even from an outside thread in the event that the thread pool already had zero threads.

That said, it would probably be nice to get rid of the redundant work...

carterkozak: I see, should the termination task be wrapped with the `classLoaderPreservingTaskUnchecked`?

Alternatively, we could shave off a stack frame for prettier stack traces by implementing a more specialized runnable to encapsulate logic from both safeRun and ContextClassLoaderSavingRunnable.

dmlloyd: Maybe we can wrap the termination task and then just stop using `safeRun` for this purpose altogether TBH.  If we always control the TCCL then there's no reason to set/restore it...

dmlloyd: New rev pushed.

dmlloyd:
New rev uses the TCCL-preserving `Runnable` wrapper and drops the corresponding logic from `safeRun`; also it fixes cases where interrupt can get lost during shutdown.  Review with whitespace hidden for sanity. :)

PR #81, opened 2020-04-30
Consolidate task execution logic used by EnhancedQueueExecutor

carterkozak:
This is an attempt to sketch out the idea proposed in https://github.com/jbossas/jboss-threads/pull/79#discussion_r417451830. I'm not entirely convinced it's worth the trade-off, looking for feedback.

Consolidate logic from safeRun, doRunTask, and classLoaderPreservingTask
into a single Runnable wrapper owned by the EnhancedQueueExecutor.
This results in simpler stack traces with two fewer frames.

This complicates handoff, requiring wrapped Runnables to be
unwrapped into a form that's not directly associated with
EQE state.

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -746,7 +746,7 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
      */
     public void execute(Runnable runnable) {
         Assert.checkNotNullParam("runnable", runnable);
-        final Runnable realRunnable = JBossExecutors.classLoaderPreservingTaskUnchecked(runnable);
+        final Task realRunnable = new Task(runnable);
         int result;
         result = tryExecute(realRunnable);
         boolean ok = false;
@@ -806,7 +806,7 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
                     head.setNextOrdered(head);
                     if (! NO_QUEUE_LIMIT) decreaseQueueSize();
                     head = taskNode;
-                    list.add(taskNode.task);
+                    list.add(taskNode.task.handoff());
                 }
                 // retry
             } else {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -2427,4 +2405,59 @@ public long getSpinMissCount() {
...
+
+    final class Task implements Runnable {
+
+        private final Runnable delegate;
+        private final ClassLoader contextClassLoader;
+
+        Task(Runnable delegate) {
+            Assert.checkNotNullParam("delegate", delegate);
+            this.delegate = delegate;
+            this.contextClassLoader = JBossExecutors.getContextClassLoader(Thread.currentThread());
```
carterkozak, line 2417:
```suggestion
            this.contextClassLoader = JBossExecutors.getContextClassLoaderUnchecked(Thread.currentThread());
```

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -806,7 +806,7 @@ public void shutdown() {
                     head.setNextOrdered(head);
                     if (! NO_QUEUE_LIMIT) decreaseQueueSize();
                     head = taskNode;
-                    list.add(taskNode.task);
+                    list.add(taskNode.task.handoff());
```
dmlloyd, line 809: When `shutdownNow` is called, one of two caller behaviors is likely:
* They will run the tasks
* They will discard the tasks

I think in both of these cases it's better for the task to have the correct TCCL; I think it's probably OK to also update the counters in the event that the task is actually run (better late than never).  WDYT?

carterkozak: > I think in both of these cases it's better for the task to have the correct TCCL

You are correct, the `handoff()` method does retain the tccl by wrapping the delegate task with the general purpose `ContextClassLoaderSavingRunnable`. I'm not sure we want to update counters for this executor when tasks are executed elsewhere, I don't think it should impact the shutdownNow EQE based on whether tasks are discarded or executed.

dmlloyd (approved):

PR #82, opened 2020-05-05
Remove onSpinMisses when failed due to stale view of world

franz1981:
In both cases awaiting (in the form of `onSpinMisses`) while catching up with the state of the world isn't a good idea. Probably it coud alleviate a bit the contention under load, improving artificially troughput. but it can harm max latencies. 

A good proof of this is on https://github.com/AdoptOpenJDK/openjdk-jdk14/blob/f7165c322a6abefa34afa9eeb76dbd56f70aae09/src/java.base/share/classes/java/util/concurrent/ConcurrentLinkedQueue.java#L369 (the base algorithm is very similar, based on the Michael-Scott queue): on a failed CAS due to contention or a stale view of the world, it retries immediately despite the presence of `onSpinWait` on the jdk 14 release.

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -1724,8 +1724,6 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
             tailNext = tail.getNext();
             // tail is already consumed
             if (tailNext == tail) {
-                if (UPDATE_STATISTICS) spinMisses.increment();
-                JDKSpecific.onSpinWait();
                 // retry with new tail(snapshot)
                 tail = this.tail;
                 continue;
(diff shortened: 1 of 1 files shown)


franz1981:
@carterkozak Sadly I don't have any box available ATM to verify this, so will left you too (@dmlloyd ) to decide/evaluate if this is causing evident regressions.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1724,8 +1724,6 @@ private int tryExecute(final Runnable runnable) {
             tailNext = tail.getNext();
             // tail is already consumed
             if (tailNext == tail) {
-                if (UPDATE_STATISTICS) spinMisses.increment();
```
carterkozak, line 1727: I think we still want to increment spin misses in these cases, right?

carterkozak:
Happy to run some benchmarks this evening :-)

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
franz1981: Not sure TBH: is a misses because the we've been too fast or we've bee to slow? I suppose the latter..I don't know if we should consider both cases as "spin misses" or maybe yes...

franz1981:
@carterkozak thanks! Probably will get a machine by the end of the week ^^

carterkozak:
I think this change is included in #84 which balances the loop structure a bit.

carterkozak:
@franz1981 I think this can be closed now that #84 has merged. Thanks for your help on that change! :-)

PR #83, opened 2020-05-05
Implemented pinned threads low(er)latency option

franz1981:


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1732,6 +1737,13 @@ private int tryExecute(final Runnable runnable) {
             }
             assert tailNext != tail;
             if (tailNext instanceof TaskNode) {
+                // we trust the producer moving tail won't be descheduled
+                // hence there is no need to help it
+                if (UNDER_SUBSCRIBED) {
+                    tail = this.tail;
```
carterkozak, line 1743: In this case do you think we should increment spin misses and call `onSpinWait` since we may be waiting for another thread to complete an action?

franz1981: I really don't have any idea, maybe yes, but I'm not sure it has the same meaning with threads < cores because we always performs offers in 2 steps and given that we're not allowed to "help" the active producer is a "common" scenario that we can see an operation in an "half-completed" state...Just thinking loud...

carterkozak:
I've uploaded both the table and json here:
https://gist.github.com/carterkozak/8c0b6cd852a4c7a820b577679a5850bb

All of the benchmarks regardless of the configured core count use 32 threads to submit tasks, which may be pushing it on a 14 core (28 with smt) machine.

franz1981:
> All of the benchmarks regardless of the configured core count use 32 threads to submit tasks, which may be pushing it on a 14 core (28 with smt) machine.

The change I have performed here should be enabled when the total threads <= cores, I don't know if we can have benefit in other cases

carterkozak:
with 7 producers and 7 pooled threads there's not a great deal of difference between the two modes (assuming I've set up the benchmark properly):

```
Benchmark                           (burstSize)  (cores)  (executorThreads)                                 (factory)  (spinWaitCompletion)  Mode  Cnt    Score   Error  Units
ExecutorBenchmarks.benchmarkSubmit            1       28                7,7                  LTQ_THREAD_POOL_EXECUTOR                  true  avgt    4    2.866 ± 0.199  us/op
ExecutorBenchmarks.benchmarkSubmit            1       28                7,7                  LBQ_THREAD_POOL_EXECUTOR                  true  avgt    4   10.027 ± 1.340  us/op
ExecutorBenchmarks.benchmarkSubmit            1       28                7,7                   ENHANCED_QUEUE_EXECUTOR                  true  avgt    4    3.730 ± 0.043  us/op
ExecutorBenchmarks.benchmarkSubmit            1       28                7,7  ENHANCED_QUEUE_EXECUTOR_UNDER_SUBSCRIBED                  true  avgt    4    3.770 ± 0.039  us/op
ExecutorBenchmarks.benchmarkSubmit           10       28                7,7                  LTQ_THREAD_POOL_EXECUTOR                  true  avgt    4   22.489 ± 0.509  us/op
ExecutorBenchmarks.benchmarkSubmit           10       28                7,7                  LBQ_THREAD_POOL_EXECUTOR                  true  avgt    4   38.257 ± 5.157  us/op
ExecutorBenchmarks.benchmarkSubmit           10       28                7,7                   ENHANCED_QUEUE_EXECUTOR                  true  avgt    4   40.510 ± 0.642  us/op
ExecutorBenchmarks.benchmarkSubmit           10       28                7,7  ENHANCED_QUEUE_EXECUTOR_UNDER_SUBSCRIBED                  true  avgt    4   40.040 ± 0.208  us/op
ExecutorBenchmarks.benchmarkSubmit          100       28                7,7                  LTQ_THREAD_POOL_EXECUTOR                  true  avgt    4  242.368 ± 1.920  us/op
ExecutorBenchmarks.benchmarkSubmit          100       28                7,7                  LBQ_THREAD_POOL_EXECUTOR                  true  avgt    4  260.482 ± 2.755  us/op
ExecutorBenchmarks.benchmarkSubmit          100       28                7,7                   ENHANCED_QUEUE_EXECUTOR                  true  avgt    4  397.269 ± 5.073  us/op
ExecutorBenchmarks.benchmarkSubmit          100       28                7,7  ENHANCED_QUEUE_EXECUTOR_UNDER_SUBSCRIBED                  true  avgt    4  412.037 ± 3.659  us/op
```

The profiler shows nearly 20% of samples in compareAndSetNext, which I imagine dominates changes to tail.

franz1981:
@carterkozak That's very important & interesting so the optimization performed by LTQ (the same of the latest CLQ) that allows to perform 2 jumps past tail to save awaiting it to be set (and saving to set it too half of the times) is much more important to improve the behaviour under contention then just lowering the cost of submission (that probably could benefit just the single-threaded case at this point)

franz1981:
At this point I can close this, because probably the real deal would be to perform such optimization instead of this one (concurrency utils indeed are a awesome piece of engineering!)

PR #84, opened 2020-05-05
Refactor the tryExecute loop

carterkozak:
The major change this provides is removal of `onSpinWait` from
spin-miss paths because in these cases there's nothing to wait for:
a race has already been lost.

Secondly, this reduces duplicate code in the spin-miss path
and consolidates spin-miss code to the end of the loop after
conditionals have been chekced without returning.

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -1722,31 +1722,55 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
         TaskNode node = null;
         for (;;) {
             tailNext = tail.getNext();
-            // tail is already consumed
             if (tailNext == tail) {
-                if (UPDATE_STATISTICS) spinMisses.increment();
-                JDKSpecific.onSpinWait();
-                // retry with new tail(snapshot)
-                tail = this.tail;
-                continue;
-            }
-            assert tailNext != tail;
-            if (tailNext instanceof TaskNode) {
-                TaskNode tailNextTaskNode = (TaskNode) tailNext;
-                // Opportunistically update tail to the next node. If this operation has been handled by
-                // another thread we fall back to the loop and try again instead of duplicating effort.
-                if (!compareAndSetTail(tail, tailNextTaskNode)) {
-                    if (UPDATE_STATISTICS) spinMisses.increment();
-                    JDKSpecific.onSpinWait();
-                    tail = this.tail;
-                    continue;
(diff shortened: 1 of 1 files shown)


carterkozak:
I haven't run benchmarks against this yet, I'll report back with findings once I have them.

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1722,30 +1722,6 @@ private int tryExecute(final Runnable runnable) {
         TaskNode node = null;
         for (;;) {
             tailNext = tail.getNext();
-            // tail is already consumed
```
franz1981, line 1725: I would left this check to be executed as the first one, because checking `instance of` force walking the next node until reading the class word, while ref check nope

carterkozak: That's a great point, I'll update :-)

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1784,9 +1762,6 @@ private int tryExecute(final Runnable runnable) {
                     }
                     // otherwise the consumer gave up or was exited already, so fall out and...
                 }
-                if (UPDATE_STATISTICS) spinMisses.increment();
-                // retry with new tail(snapshot) as was foretold
-                tail = this.tail;
             } else if (tailNext == null) {
```
carterkozak, line 1765: We could prioritize the null check as well, but I think this case is rare. There are no available threads and the queue is empty.

franz1981: If the queue is empty for real, the instanceof won't fetch the class word because the null check is performed earlier. The interesting part is that we save it to be performed twice probably (not sure)

carterkozak: That's a good point, the earlier null check may produce simpler optimized code. I'll reorder this later today.

carterkozak:
Results from last night, I haven't had a chance to dig through them in depth but they appear to be within the margin of error for existing results with a few showing possible improvement:
https://gist.github.com/carterkozak/dd51dd09bc005ed20f714a4064b4f92c

dmlloyd:
Am I correct that this does not contain any functional changes (other than removing onSpinWait in some paths), otherwise simply reordering the loop conditionals?

carterkozak:
You are correct.

dmlloyd (approved):

PR #85, opened 2020-05-14
Implement CachedExecutorView to split underlying cached pools

carterkozak:
This allows us to provide references to the same underlying pool of
threads to different consumers and utilize distinct thread name patterns
and instrumentation without duplicating resources.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,6 +110,18 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.16.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>4.0.2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.jboss.logmanager</groupId>
             <artifactId>jboss-logmanager</artifactId>
             <version>2.1.14.Final</version>
--- a/src/main/java/org/jboss/threads/QueuedViewExecutor.java
+++ b/src/main/java/org/jboss/threads/QueuedViewExecutor.java
@@ -1,0 +1,288 @@
+package org.jboss.threads;
+
+import org.jboss.logging.Logger;
+import org.wildfly.common.Assert;
+import org.wildfly.common.lock.ExtendedLock;
+import org.wildfly.common.lock.Locks;
+
+import java.util.ArrayDeque;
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -109,6 +109,18 @@
...
+            <artifactId>assertj-core</artifactId>
+            <version>3.16.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>4.0.2</version>
+            <scope>test</scope>
+        </dependency>
```
carterkozak, line 123: I left these in from my original tests, the tests can be rewritten to use standard junit assert functionality instead of third party libraries, I didn't want to block creating the initial PR on it.

src/main/java/org/jboss/threads/CachedExecutorView.java
```diff
@@ -0,0 +1,298 @@
...
+
+        @Override
+        public void run() {
+            Thread currentThread = Thread.currentThread();
+            Set<CachedExecutorViewRunnable> runnables = activeRunnables;
+            this.thread = currentThread;
+
+            String originalName = currentThread.getName();
+            try {
+                currentThread.setName(nameFunction.rename(originalName));
```
carterkozak, line 208: Rename doesn't need to be called each invocation, a threadlocal anchored to the CachedExecutorView instance should allow us to calculate these once per thread.

carterkozak: Implementations of the name function could implement the threadlocal piece themselves as well, that way a an identity thread name function doesn't incur the cost of a map lookup.

src/main/java/org/jboss/threads/CachedExecutorView.java
```diff
@@ -0,0 +1,298 @@
...
+ * View of a cached executor service which avoids lock contention in the common path. This allows us to
+ * provide references to the same underlying pool of threads to different consumers and utilize distinct
+ * thread name patterns and instrumentation without duplicating resources. We take advantage of the fact that
+ * cached executors never queue, and always submit work immediately to avoid implementing queueing.
+ *
+ * @author <a href="mailto:ckozak@ckozak.net">Carter Kozak</a>
+ */
+final class CachedExecutorView extends AbstractExecutorService {
+    private static final AtomicIntegerFieldUpdater<CachedExecutorView> stateUpdater =
+            AtomicIntegerFieldUpdater.newUpdater(CachedExecutorView.class, "state");
```
carterkozak, line 46: I'll plan to replace this with unsafe for a bit of additional optimization and uniformity with the rest of the project.

PR #86, opened 2020-06-15
JBTHR-86: Fix interruption race in ViewExecutor

carterkozak:
https://issues.redhat.com/browse/JBTHR-86

### DIFF
--- a/src/main/java/org/jboss/threads/QueuedViewExecutor.java
+++ b/src/main/java/org/jboss/threads/QueuedViewExecutor.java
@@ -1,7 +1,6 @@
 package org.jboss.threads;
 
 import org.jboss.logging.Logger;
-import org.wildfly.common.Assert;
 import org.wildfly.common.lock.ExtendedLock;
 import org.wildfly.common.lock.Locks;
 
@@ -10,7 +9,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.AbstractExecutorService;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.RejectedExecutionException;
@@ -213,7 +211,7 @@ final class QueuedViewExecutor extends ViewExecutor {
             this.command = command;
         }
 
-        void interrupt() {
+        synchronized void interrupt() {
             final Thread thread = this.thread;
             if (thread != null) {
                 thread.interrupt();
@@ -221,7 +219,7 @@ final class QueuedViewExecutor extends ViewExecutor {
         }
 
         public void run() {
-            boolean removeFromAllWrappersAfterCompletion = true;
+            boolean resetStateOnCompletion = true;
             thread = Thread.currentThread();
             try {
                 for (;;) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jboss/threads/ViewExecutor.java
```diff
@@ -361,7 +361,11 @@ public void run() {
                     }
                 }
             } finally {
-                thread = null;
+                // Must be synchronized with interrupt() to avoid acquiring the thread reference as work completes
+                // and interrupting the next task run by this thread which may not originate from this view.
+                synchronized (this) {
+                    thread = null;
```
carterkozak, line 367: Note that in the case this TaskWrapper re-submits itself (line 354) it's possible execution has already begun on another thread, and this may null out that thread value.

carterkozak: updated to resolve this case as well.

src/main/java/org/jboss/threads/ViewExecutor.java
```diff
@@ -307,13 +306,17 @@ void interrupt() {
...
             thread = Thread.currentThread();
             try {
                 for (;;) {
                     lock.lock();
                     try {
                         submittedCount--;
                         runningCount++;
+                        // Interruption may be missed between when a TaskWrapper is submitted
+                        // to the delegate executor, and when the task begins to execute.
+                        if (state == ST_SHUTDOWN_INT_REQ) Thread.currentThread().interrupt();
```
carterkozak, line 319: Added one more fix when I rebased this change. There's some time which wasn't accounted for between submitting a tack wrapper to the underlying executor and TaskWrapper.run being invoked where we could fail to interrupt the thread.

carterkozak: Moving this just after the `thread` field is is set, there's no reason to do this under a lock.

src/main/java/org/jboss/threads/QueuedViewExecutor.java
```diff
@@ -213,16 +213,20 @@ public String toString() {
...
         }
 
         public void run() {
-            boolean removeFromAllWrappersAfterCompletion = true;
+            boolean resetStateOnCompletion = true;
             thread = Thread.currentThread();
+            // Interruption may be missed between when a TaskWrapper is submitted
+            // to the delegate executor, and when the task begins to execute.
+            // This must execute after thread is set.
+            if (state == ST_SHUTDOWN_INT_REQ) Thread.currentThread().interrupt();
```
carterkozak, line 229: This would be more correct inside the retry loop. If the delegate has shut down and the view is terminated, it's possible a previous interruption was handled by a previous `command` in this wrapper, and must be set again for subsequent tasks.

src/main/java/org/jboss/threads/QueuedViewExecutor.java
```diff
@@ -213,18 +211,26 @@ public String toString() {
...
             thread = Thread.currentThread();
             try {
                 for (;;) {
+                    // Interruption may be missed between when a TaskWrapper is submitted
+                    // to the delegate executor, and when the task begins to execute.
+                    // This must execute after thread is set.
+                    // Must occur in the retry loop to set interruption when requested
+                    // after a command has consumed an initial interruption and the
+                    // delegate has rejected the next command.
+                    if (state == ST_SHUTDOWN_INT_REQ) Thread.currentThread().interrupt();
```
carterkozak, line 232: `state` is not volatile, this must occur under a lock.

src/test/java/org/jboss/threads/QueuelessViewExecutorTest.java
```diff
@@ -80,7 +123,7 @@ public void testShutdownNow() throws InterruptedException {
     public void testShutdownNow_immediatelyAfterTaskIsSubmitted() throws InterruptedException {
```
carterkozak, line 123: This test would fail on the queued implementation without the fix provided by this PR. The queueless implementation already does what I've proposed for the queued implementation.

PR #89, opened 2020-06-16
[JBTHR-88] Thread interruption fix

dmlloyd:
Fix problem where thread interruption from other thread can hide interruption from same thread

### DIFF
--- a/src/main/java/org/jboss/threads/JBossThread.java
+++ b/src/main/java/org/jboss/threads/JBossThread.java
@@ -18,6 +18,7 @@
 
 package org.jboss.threads;
 
+import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
@@ -27,6 +28,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.LockSupport;
 
 import org.wildfly.common.Assert;
+import org.wildfly.common.cpu.ProcessorInfo;
 import org.wildfly.common.function.ExceptionBiConsumer;
 import org.wildfly.common.function.ExceptionBiFunction;
 import org.wildfly.common.function.ExceptionConsumer;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/threads/JBossThread.java
```diff
@@ -153,18 +153,27 @@ public void interrupt() {
...
+            if (oldVal == STATE_INTERRUPT_PENDING) {
                 // already set
                 Messages.msg.tracef("Interrupting thread \"%s\" (already interrupted)", this);
                 return;
-            } else if (oldVal == STATE_INTERRUPT_DEFERRED) {
-                newVal = STATE_INTERRUPT_PENDING;
+            } else if (oldVal == STATE_INTERRUPT_IN_PROGRESS) {
+                // wait for interruption on other thread to be completed
+                JDKSpecific.onSpinWait();
+                continue;
```
yersan, line 165: @dmlloyd What about JDK 8 with only one core? Could it be an endless loop?

dmlloyd: Not endless, because the thread could be rescheduled at some point.  But this is the most naive possible spin loop implementation.  A more robust loop would spin *n* times before falling back to `Thread.yield()`, where *n* is a constant number, but is 0 on single-core systems.  Since the fix seems to work, I'll update it accordingly.

src/main/java/org/jboss/threads/JBossThread.java
```diff
@@ -153,18 +160,32 @@ public void interrupt() {
...
-            if (oldVal == STATE_INTERRUPT_PENDING || oldVal == STATE_INTERRUPT_IN_PROGRESS) {
+            if (oldVal == STATE_INTERRUPT_PENDING) {
                 // already set
                 Messages.msg.tracef("Interrupting thread \"%s\" (already interrupted)", this);
                 return;
-            } else if (oldVal == STATE_INTERRUPT_DEFERRED) {
-                newVal = STATE_INTERRUPT_PENDING;
+            } else if (oldVal == STATE_INTERRUPT_IN_PROGRESS) {
+                // wait for interruption on other thread to be completed
+                if (spins < MAX_INTERRUPT_SPINS) {
```
yersan, line 172: spin is not autoincremented on each loop, missed a spin++?

dmlloyd: You're right; fixed.

PR #91, opened 2020-06-30
Improve performance of execute under contention by using flat combiners

franz1981:
This is a kind of POC to improve performance of execute under heavy contention without modifying the original algorithm and assumptions. 

@carterkozak @dmlloyd Feel free to both test and ask on the Zulipchat channel!!!

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -755,7 +755,13 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
         final int result;
         try (TryExecuteTask tryExecuteTask = POOLED_COMBINER_TASKS.get().on(this, realRunnable)) {
             tryExecuteCombiner.combine(tryExecuteTask, idleCount -> {
-                JDKSpecific.onSpinWait();
+                final int spinWait = PARK_SPINS - YIELD_FACTOR;
+                if (idleCount >= spinWait) {
+                    Thread.yield();
+                } else {
+                    JDKSpecific.onSpinWait();
+                    idleCount++;
+                }
                 return idleCount;
             });
             result = tryExecuteTask.result;
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -50,6 +50,7 @@ import java.util.concurrent.locks.LockSupport;
 import javax.management.ObjectInstance;
 import javax.management.ObjectName;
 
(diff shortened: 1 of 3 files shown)


carterkozak:
I'm having a busy week, but I'm planning to read through this when I have time!

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -747,8 +752,20 @@ public Builder setMBeanName(final String mBeanName) {
...
+            tryExecuteCombiner.combine(tryExecuteTask, idleCount -> {
+                final int spinWait = PARK_SPINS - YIELD_FACTOR;
+                if (idleCount >= spinWait) {
+                    Thread.yield();
+                } else {
+                    JDKSpecific.onSpinWait();
+                    idleCount++;
+                }
+                return idleCount;
+            });
```
carterkozak, line 766: I think we can extract this IdleStrategy to a singleton and avoid allocation cost.

src/main/java/org/jboss/threads/combiner/Combiner.java
```diff
@@ -0,0 +1,54 @@
...
+        int idle(int idleCount);
+
+        static IdleStrategy busySpin() {
+            return ignore -> ignore;
+        }
+
+        static IdleStrategy yield(int maxSpins) {
+            return idleCount -> {
+                if (idleCount < maxSpins) {
+                    idleCount++;
```
carterkozak, line 33: Should these use onSpinWait? I think they'd need to move into the `org.jboss.threads` package for access.

carterkozak: I don't think these factory methods are used, should they be?

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -747,8 +752,20 @@ public Builder setMBeanName(final String mBeanName) {
     public void execute(Runnable runnable) {
         Assert.checkNotNullParam("runnable", runnable);
         final Runnable realRunnable = JBossExecutors.classLoaderPreservingTaskUnchecked(runnable);
-        int result;
-        result = tryExecute(realRunnable);
+        final int result;
+        try (TryExecuteTask tryExecuteTask = POOLED_COMBINER_TASKS.get().on(this, realRunnable)) {
```
carterkozak, line 756: I wonder if something like #81 would allow us to avoid this threadlocal in favor of a couple additional references on an existing object?

carterkozak:
I still don't fully understand the implementation, I'll need to stare at the code a bit longer.
My benchmark results are all over the place, I think there must be something wrong on my end, or possibly a race that can result in excessive link traversal, but it doesn't happen every run.

PR #92, opened 2020-07-07
[JBTHR-55] Thread exit hooks

msfm:
Backport https://issues.redhat.com/browse/JBTHR-55 to 2.3 branch
Upstream PR: #38

### DIFF
--- a/src/main/java/org/jboss/threads/JBossThread.java
+++ b/src/main/java/org/jboss/threads/JBossThread.java
@@ -21,6 +21,8 @@ package org.jboss.threads;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.security.PrivilegedExceptionAction;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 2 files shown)


dmlloyd:
Normally we don't backport features to minor branches (otherwise the minor branch eventually becomes a more complex and harder to maintain copy of a newer branch).  Can we use a newer version instead?

wolfc:
WildFly (Core) depends on the API which was stripped down through https://issues.redhat.com/browse/JBTHR-62. It is not trivial to refactor it (let alone EAP 7.3).

dmlloyd:
@wolfc the stripping down occurred in 3.x, not 2.x.  I'm talking about starting a new 2.x branch (2.4.x) for this purpose, which includes all the 2.3 changes plus any additional features that need adding to get past the current problem.

wolfc:
If 2.4 becomes a simple drop-in upgrade then I see no problem going down that route.

msfm:
As using 2.4 looks better than backporting to 2.3, I'm closing this PR

PR #95, opened 2020-07-08
Implementation of a performant queueing view executor

carterkozak:
Benchmark results, run with 32 benchmark threads on Xeon W-2175 on commit 5bba76b8e07314a35d122663a68a9498155065e9

The EQE executor is included to provide a performance baseline for an isolated executor.

Burst size ten, 32 submitting threads, with 32 allowed concurrent tasks. This is meant to stress the queued implementation, thus the queueless implementation isn't comparable:
```
Benchmark                           (burstSize)  (cores)  (executorThreads)  (factory)  (spinWaitCompletion)  Mode  Cnt    Score    Error  Units
ExecutorBenchmarks.benchmarkSubmit           10       28                 32     QUEUED                 false  avgt    4  413.682 ± 36.491  us/op
ExecutorBenchmarks.benchmarkSubmit           10       28                 32   ENHANCED                 false  avgt    4  149.045 ±  0.557  us/op
ExecutorBenchmarks.benchmarkSubmit           10       28                 32        EQE                 false  avgt    4  155.282 ±  2.048  us/op
```
Note the slightly better performance using the enhanced wrapper, I think we're able to saturate the executor and rely on the ConcurrentLinkedQueue which includes optimizations @franz1981 kindly described.

Benchmarks without queueing, the workload the queueless implementation was built for. The 100 executor threads isn't meaningful by itself, it's enough permits that the queue isn't used, threads are bounded by the 32 benchmarking threads which submit one task at a time. In some runs the queueless implementation semes to slightly outperform  the enhanced version, but not by enough that it's worthwhile to keep both around.
```
Benchmark                           (burstSize)  (cores)  (executorThreads)  (factory)  (spinWaitCompletion)  Mode  Cnt    Score   Error  Units
ExecutorBenchmarks.benchmarkSubmit            1       28                100     QUEUED                 false  avgt    4  127.395 ± 6.362  us/op
ExecutorBenchmarks.benchmarkSubmit            1       28                100  QUEUELESS                 false  avgt    4   17.152 ± 1.045  us/op
ExecutorBenchmarks.benchmarkSubmit            1       28                100   ENHANCED                 false  avgt    4   17.101 ± 1.090  us/op
ExecutorBenchmarks.benchmarkSubmit            1       28                100        EQE                 false  avgt    4   16.059 ± 5.820  us/op
```


carterkozak:
This implementation performs similarly to the queueless view when the queue is not used (either disabled, or tasks don't fill available permits). In heavily contended benchmarks it shows up to ~3% slower performance compared to the queueless implementation, which is close enough that I'd consider them equivalent. Benchmarks against the queued implementation show a significant improvement, even in scenarios which are built to cause frequent use of the executeLock. Improvement ranges from 100% to 1000% (in cases where the lock degraded performance forcing waiters to yield)

src/main/java/org/jboss/threads/ViewExecutor.java
```diff
@@ -100,6 +105,14 @@ public Builder setQueueInitialSize(final int queueInitialSize) {
...
+            if (USE_V2) {
+                // queueInitialSize is not relevant using ConcurrentLinkedQueue
+                return new EnhancedViewExecutor(
+                        Assert.checkNotNullParam("delegate", delegate),
+                        maxSize,
+                        queueLimit,
+                        handler);
+            }
             if (queueLimit == 0) {
                 return new QueuelessViewExecutor(Assert.checkNotNullParam("delegate", delegate), maxSize, handler);
```
carterkozak, line 117: The QueuelessViewExecutor can safely be deleted since it has not been released and the EnhancedViewExecutor performs just as well.

Depending on risk tolerance the QueuedViewExecutor and feature flag could be removed as well. I defer to your judgement.

dmlloyd: Let's definitely delete Queueless.  Less code means fewer bugs.

carterkozak: Sounds good. Do you have any preference on the default value of the `org.jboss.threads.view-executor.v2` flag to use the enhanced version when a queue is requested?
If we leave it enabled by default, I'll add javadoc and a deprecation to the get/setQueueInitialSize methods.

carterkozak:
I've added changes as separate commits to avoid causing confusion in review, I'll squash them together once we're happy with the change.

src/main/java/org/jboss/threads/ViewExecutor.java
dmlloyd: Let's just have one implementation; we don't need to keep the old one around IMO.  I think it's probably fine to deprecate those methods.

carterkozak: Sounds reasonable. I've deleted the QueuedViewExecutor implementation.

It may be helpful to add padding to prevent noisy neighbor problems, wdyt? Perhaps not necessary for this PR.

dmlloyd (approved):

dmlloyd:
Thanks!

PR #98, opened 2020-07-13
Classloader preservation is configurable

carterkozak:
This updates QueuedViewExecutor to use the unchecked variation of
the task factory, and adds support for classloader preservation
to the queueless implementation.

This enhances the EQE builder to support configurable classloader
preservation in order to avoid wrapping tasks multiple times when
an EQE is used as a base executor and only used through views.


carterkozak:
closing in favor of simply adding support to the cached version

PR #100, opened 2020-09-28
Remove the ViewExecutor artificial size limit

carterkozak:
Unlike the previous implementations, EnhancedViewExecutor stores
state in a 64-bit integer, allowing both the pool and queue
sizes to use full integers.

There's not any reason to create an executor with more than 32k
active tasks, however the builder uses the `int` type so it's nice
to allow `Integer.MAX_VALUE` to have effectively the same impact
as `Short.MAX_VALUE` and rely on the compiler for input validation.


carterkozak:
Very low priority, I just realized I should have removed the cast and assert when I removed the old ViewExecutor implementations.

dmlloyd (approved):

PR #102, opened 2020-11-13
Display the version in DEBUG level

gastaldi:
Showing the version pollutes the log and it's pretty useless unless you're well, debugging :)

/cc @dmlloyd

### DIFF
--- a/src/main/java/org/jboss/threads/Messages.java
+++ b/src/main/java/org/jboss/threads/Messages.java
@@ -39,7 +39,7 @@ interface Messages extends BasicLogger {
 
     // version
     @Message(value = "JBoss Threads version %s")
-    @LogMessage(level = Logger.Level.INFO)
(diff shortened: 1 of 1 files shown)


dmlloyd:
This is something that was requested by support and thus enacted into policy many years ago.  The user log would always show all of the versions in use, allowing for more accurate troubleshooting.  AFAIK this policy is still in place for the entire JBoss product line, so this should not be changed until that policy is retracted or replaced with something else (for example, using class loaders or JPMS modules that in turn can add module names and versions into the stack traces or log in Java 9+).  This can be revisited with support to see if the policy can be changed or retired though.

galderz:
@dmlloyd Can we integrate this change once and for all? Seems like JBoss Threads is the only one out there, see https://github.com/quarkusio/quarkus/issues/23748

dmlloyd:
Resolved by #120.

PR #106, opened 2021-05-12
Simple context propagation for e.g. vert.x

dmlloyd:


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -404,6 +416,8 @@ public ObjectInstance run() {
         private int maxQueueSize = Integer.MAX_VALUE;
         private boolean registerMBean = REGISTER_MBEAN;
         private String mBeanName;
+        private Supplier<Object> contextSupplier = () -> null;
+        private Consumer<Object> contextConsumer = obj -> {};
```
carterkozak, line 420: I'd be keen to use this to support tracing without additional ExecutorService wrapping or stack frames.

Thoughts on combining these into a single interface with a generic type?
e.g.
```java
public interface ExecutorContextHandler<T> {
    T createFromSubmittingThread();
    void applyContextToExecutorThread(T context);
    // perhaps this could also use some way to clear the context after a task completes?
    void clearContext();
}
```
My example method names are terrible, but I think this captures the idea.

dmlloyd: There isn't a convenient way (API-wise) to genericize these, hence going with Object.  I went with the simple functional objects so that they could be configured with static method refs, but I'm not married to it (maybe @kenfinnigan would like to offer an opinion as I wrote this to help him solve a Vert.x issue).  Maybe it's worth having an explicit clear as well; it depends on what the context is I suppose, and how detrimental it would be to leak it.

kenfinnigan: I'm currently playing with this and may need to "tweak" this PR based on what I find.

Once I have something doing what is needed, we can then revisit better approaches and alternatives

dmlloyd:
I have no idea why CI didn't run for this.

carterkozak:
> I have no idea why CI didn't run for this.

I think it's the branch rename:
https://github.com/jbossas/jboss-threads/pull/107

dmlloyd:
Thanks @carterkozak!

stuartwdouglas:
This would also need some way to clean up the context after the task is run.

kenfinnigan:
So I have most of this working for what I needed, but I can't cleanup the context on the `VertxThread` with a `Runnable`.

We'd need to modify the `contextConsumer` to be a `Function` that accepts the context we want to set and returns the previous one on the thread. Then `contextCleanup` would be a `Consumer<Object>` that takes the result of `contextConsumer` as input.

Happy to make the changes if everyone's ok with it

dmlloyd:
The thread pool thread doesn't have any nested context at all.  In other words, the first task always has `null` as a previous context, and since it would therefore clear the context back to `null` again at the end of the task, it is proven that the previous context will always be `null`.  That's why I figured a `Runnable` would suffice.

kenfinnigan:
Interesting. So we don't need any cleanup then if we don't need to remove the context off the VertxThread?

dmlloyd:
I think we do need a cleanup, but the cleanup can just null it out.

kenfinnigan:
Ah ok. I think I can do that with a plain Runnable, but will verify

carterkozak:
Sorry to butt in, feel free to ignore my rambling!

A design along these lines would allow nulling cleanup via a singleton `ContextCleaner`, while allowing an implementation to reset the context to a previous value at the cost of an `ContextCleaner` implementation object allocation: https://github.com/jbossas/jboss-threads/commit/656823513de4233f3bd7d0febdb6ef2520e78ce8

The wrapper executor is largely an example, but it would apply more broadly than directly to the EQE. It's an additional Runnable wrapper allocation, but may allow easier prototyping in the immediate term, and would be easier to adapt to support ScheduledExecutorServices. I admit I'm biased because the majority of the executors I use are ViewExecutor wrappers around a singleton EnhancedQueueExecutor (poor-mans-loom) so I'd need another way to take advantage of this.

I find the individual methods difficult to understand at a glance, particularly that it's safe to cast the Object value into the type produced by the supplier, where a singleton enum implementation of an interface with a type parameter provides compile time safety. That said, there may be something you're optimizing for which I don't understand.

Given the context classloader support is already built into jboss executors, it makes a good example for this structure including both types of cleanup functionality:

```java
public enum ContextClassLoaderContextHandler implements ContextExecutorService.ContextHandler<ClassLoader> {
    INSTANCE;

    @Override
    public ClassLoader collectContext() {
        return JBossExecutors.getContextClassLoader(Thread.currentThread());
    }

    @Override
    public ContextExecutorService.ContextCleaner installContext(ClassLoader context) {
        ClassLoader original = JBossExecutors.getAndSetContextClassLoader(Thread.currentThread(), context);
        if (original == context) {
            return null;
        }
        if (original == null) {
            return ContextClassLoaderRestoringContextCleaner.NULL_CCL;
        }
        return new ContextClassLoaderRestoringContextCleaner(original);
    }

    private static final class ContextClassLoaderRestoringContextCleaner implements ContextExecutorService.ContextCleaner {
        private static final ContextClassLoaderRestoringContextCleaner NULL_CCL = new ContextClassLoaderRestoringContextCleaner(null);

        private final ClassLoader original;

        ContextClassLoaderRestoringContextCleaner(ClassLoader original) {
            this.original = original;
        }

        @Override
        public void close() {
            JBossExecutors.setContextClassLoader(Thread.currentThread(), original);
        }
    }
}
```

kenfinnigan:
I've confirmed that the cleanup taking a `Runnable` I can make work by utilizing the same method the `contextSupplier` uses to retrieve the current Context and then call `endDispatch` with a null context as the replacement.

This PR is all good from my side

kenfinnigan (approved):
Looks good to me

src/main/java/org/jboss/threads/ContextHandler.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * A handler for propagating context from a task submitter to a task execution.
+ *
+ * @param <T> the class of the context to capture
+ */
+public interface ContextHandler<T> {
+    /**
+     * The context handler which captures no context.
+     */
+    ContextHandler<?> NONE = new ContextHandler<Object>() {
```
carterkozak, line 12: We could specify `ContextHandler<Void>`, but it shouldn't make any difference

carterkozak (approved):
The API is very clean, thank you!

kenfinnigan:
To implement a cleanup I just add it after `task.run()` inside `runWith()`?

The only question I have is whether this is too much flexibility because we get the task to run, enabling something entirely different to be done, or wrapping the `Runnable` inside other bits. If that's intended and ok that's fine, wanted to check

dmlloyd:
> To implement a cleanup I just add it after `task.run()` inside `runWith()`?

Yes, but in a `finally` block.

> The only question I have is whether this is too much flexibility because we get the task to run, enabling something entirely different to be done, or wrapping the `Runnable` inside other bits. If that's intended and ok that's fine, wanted to check

This is the intent. When scope locals hit, this is the only way we can do it because they don't have a "start/clear context" concept - you can only propagate the context by passing a `Runnable` to it.

kenfinnigan:
Ah ok, so this is to cater to the JEP going through that you posted about the other day?

No problem, all good with me then

PR #108, opened 2021-07-30
Avoid capturing ACC when not necessary

Sanne:
We can avoid capturing the AccessControlContext when MBeans are disabled.

The `acc` field is holding a reference to the classloader, which is leading us to have a classloader leak in Quarkus during some integration tests - in some cases it leads to critical failures as we exhaust memory on CI.

There might be better ways to avoid the CL leak in Quarkus but this seems easy enough and useful anyway?

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -189,6 +189,7 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
     private final Object handle;
     /**
      * The access control context of the creating thread.
+     * Will be set to null when the MBean is not registered.
      */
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -1859,7 +1861,7 @@ void completeTermination() {
                 waiters = waiters.getNext();
             }
             tail.setNext(TERMINATE_COMPLETE);
-            if (! DISABLE_MBEAN) {
```
stuartwdouglas, line 1862: native-image can no longer DCE this path, which can result in failures due to JMX now being reachable.

Sanne: ouch, sorry!

PR #109, opened 2021-08-02
Set AccessControlContext to null on termination to release captured c…

Sanne:
…lassloaders

Followup to #108 as suggested by @dmlloyd

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -1861,13 +1861,17 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
                 waiters = waiters.getNext();
             }
             tail.setNext(TERMINATE_COMPLETE);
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -191,7 +191,7 @@
      * The access control context of the creating thread.
      * Will be set to null when the MBean is not registered.
      */
-    private final AccessControlContext acc;
+    private volatile AccessControlContext acc;
```
carterkozak, line 194: Is volatile necessary given that it's only updated via `completeTermination`?

Sanne: Yes I think it's better to make it `volatile`. It certainly doesn't hurt.

Sanne:
I've added also commit b3cb132e0fb4d503ca44a1bac62d0e3bf2c8c5dc to fix a problem I introduced in my previous PR.

PR #119, opened 2022-08-23
Allow silencing version logging with a system property

gsmet:
/cc @dmlloyd @yrodiere

### DIFF
--- a/src/main/java/org/jboss/threads/Version.java
+++ b/src/main/java/org/jboss/threads/Version.java
@@ -31,6 +31,9 @@ public final class Version {
     private Version() {
     }
 
+    // No specific prefix here, this controls the behavior of multiple JBoss libraries
+    private static final String JBOSS_LOG_VERSIONS = "jboss.log-versions";
+
     private static final String JAR_NAME;
     private static final String VERSION_STRING;
 
@@ -48,9 +51,17 @@ public final class Version {
         }
         JAR_NAME = jarName;
         VERSION_STRING = versionString;
+
+        boolean logVersion = true;
         try {
-            Messages.msg.version(versionString);
+            logVersion = !"false".equalsIgnoreCase(System.getProperty(JBOSS_LOG_VERSIONS));
         } catch (Throwable ignored) {}
+
+        if (logVersion) {
+            try {
+                Messages.msg.version(versionString);
+            } catch (Throwable ignored) {}
+        }
     }
 
     /**


yrodiere (approved):
LGTM

Context: https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/Disable.20version.20logging

src/main/java/org/jboss/threads/Version.java
```diff
@@ -48,9 +51,17 @@ private Version() {
         }
         JAR_NAME = jarName;
         VERSION_STRING = versionString;
+
+        boolean logVersion = true;
         try {
-            Messages.msg.version(versionString);
+            logVersion = !"false".equalsIgnoreCase(System.getProperty(JBOSS_LOG_VERSIONS));
         } catch (Throwable ignored) {}
```
gsmet, line 58: Decided to be extra defensive to keep the original behavior even if `System.getProperty` throws something.

darranl: +1 was just thinking what would happen if a security manager prevented the call.

jamezp (approved):

src/main/java/org/jboss/threads/Version.java
```diff
@@ -31,6 +31,9 @@ public final class Version {
     private Version() {
     }
 
+    // No specific prefix here, this controls the behavior of multiple JBoss libraries
+    private static final String JBOSS_LOG_VERSIONS = "jboss.log-versions";
```
gsmet, line 35: Wondering if we should make this public for convenience but I'm not sure we want to commit to it always being around.

jamezp: This is more of a convention so I'd opt for it not being public. More than likely too you'd want the property configured before you access the type as logging would need to be configured first.

gsmet: Ah ah yeah. I thought of that yesterday and then forgot today why I didn’t post the proposal in the end :).

src/main/java/org/jboss/threads/Version.java
```diff
@@ -48,9 +51,17 @@ private Version() {
         }
         JAR_NAME = jarName;
         VERSION_STRING = versionString;
+
+        boolean logVersion = true;
         try {
-            Messages.msg.version(versionString);
+            logVersion = !"false".equalsIgnoreCase(System.getProperty(JBOSS_LOG_VERSIONS));
```
dmlloyd, line 57: I would strongly prefer usage of `parseBoolean` here, and use `"true"` as the default value for the property get.

gsmet: @dmlloyd yes let’s go with your PR.

PR #120, opened 2022-08-26
Do not log versions if `jboss.log-version` is not `true`

dmlloyd:


src/main/java/org/jboss/threads/Version.java
```diff
@@ -48,7 +50,8 @@ private Version() {
         }
         JAR_NAME = jarName;
         VERSION_STRING = versionString;
-        try {
+        boolean logVersion = AccessController.doPrivileged((PrivilegedAction<Boolean>) () -> Boolean.valueOf(Boolean.parseBoolean(System.getProperty("jboss.log-version", "true")))).booleanValue();
```
gastaldi, line 53:
```suggestion
        boolean logVersion = AccessController.doPrivileged((PrivilegedAction<Boolean>) () -> Boolean.valueOf(System.getProperty("jboss.log-version", "true"))).booleanValue();
```

gsmet, line 53: FWIW, in my version (which is missing the privileged thing for sure), I put this part in a try/catch to make sure we keep the current behavior even if something wrong gets in the way:
https://github.com/jbossas/jboss-threads/pull/119/files#diff-f7179ffcce5656c32a94e972f59059723d7d9d1c53bc3274c1d7bdda0b6a212aR57

src/main/java/org/jboss/threads/Version.java
```diff
@@ -48,11 +50,20 @@ private Version() {
...
+        if (logVersion) try {
             Messages.msg.version(versionString);
         } catch (Throwable ignored) {}
     }
 
+    private static Boolean shouldLogVersion() {
+        try {
+            return Boolean.valueOf(System.getProperty("jboss.log-version", "true"));
+        } catch (Throwable ignored) {
+            return Boolean.FALSE;
```
gastaldi, line 63: Should it return `Boolean.TRUE` here?

dmlloyd: Nah, I prefer to err on the side of "less annoying". :)

src/main/java/org/jboss/threads/Version.java
```diff
@@ -48,11 +50,20 @@ private Version() {
         }
         JAR_NAME = jarName;
         VERSION_STRING = versionString;
-        try {
+        boolean logVersion = AccessController.doPrivileged((PrivilegedAction<Boolean>) Version::shouldLogVersion).booleanValue();
```
gsmet, line 53: The way you did it, if you have an exception here, you won't log anything. Which is not the preexisting behavior.

But I will let you judge of if it's a problem or not.

dmlloyd: TBH I really want the default to be `false`; anyway I prefer to err on the side of the more desirable behavior.

dmlloyd:
I rebased this on the earliest possible commit, so I can merge the same commit into stable branches.

PR #132, opened 2023-03-31
Saving setting class loader over and over again (and card-marking).

franz1981:
Similarly to https://github.com/quarkusio/quarkus/pull/30024 this can cause card marking to be set over and over (or any inter-generational/inter-regiona/GC barrier) that can both cause a scalability issue and wasteful work.

Some GC already use conditional card marking, that can help here, but still, the (useless, if it hasn't changed) restore to the `old` ClassLoader after the task has been executed can still cause cache misses, depending what the task does; saving such work will improve performance regardless what GC does with its barrier.

I've found this while running https://github.com/quarkusio/quarkus-quickstarts/tree/main/hibernate-orm-panache-quickstart :

image

In violet the `getAndSet` and restore of old class loader.

~5% throughput drop due to this (and cpu cycles spent); with just 2 cores. It increases depending the number of threads, granularity of tasks and number of cores (with G1GC and jdk 11).

### DIFF
--- a/src/main/java/org/jboss/threads/ContextClassLoaderSavingRunnable.java
+++ b/src/main/java/org/jboss/threads/ContextClassLoaderSavingRunnable.java
@@ -12,11 +12,14 @@ class ContextClassLoaderSavingRunnable implements Runnable {
 
     public void run() {
         final Thread currentThread = Thread.currentThread();
+        final ClassLoader newCl = loader;
         final ClassLoader old = JBossExecutors.getAndSetContextClassLoader(currentThread, loader);
         try {
             delegate.run();
         } finally {
-            JBossExecutors.setContextClassLoader(currentThread, old);
+            if (old != newCl) {
+                JBossExecutors.setContextClassLoader(currentThread, old);
+            }
         }
     }
 
--- a/src/main/java/org/jboss/threads/ContextClassLoaderSavingRunnable.java
+++ b/src/main/java/org/jboss/threads/ContextClassLoaderSavingRunnable.java
@@ -17,9 +17,7 @@ class ContextClassLoaderSavingRunnable implements Runnable {
         try {
             delegate.run();
         } finally {
-            if (old != newCl) {
-                JBossExecutors.setContextClassLoader(currentThread, old);
-            }
+            JBossExecutors.setContextClassLoader(currentThread, old);
         }
     }
 
--- a/src/main/java/org/jboss/threads/ContextClassLoaderSavingRunnable.java
+++ b/src/main/java/org/jboss/threads/ContextClassLoaderSavingRunnable.java
@@ -12,7 +12,6 @@ class ContextClassLoaderSavingRunnable implements Runnable {
 
     public void run() {
         final Thread currentThread = Thread.currentThread();
-        final ClassLoader newCl = loader;
         final ClassLoader old = JBossExecutors.getAndSetContextClassLoader(currentThread, loader);
         try {
             delegate.run();
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -2490,7 +2490,8 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
             }
             if (UPDATE_ACTIVE_COUNT) incrementActiveCount();
             final Thread currentThread = Thread.currentThread();
-            final ClassLoader old = JBossExecutors.getAndSetContextClassLoader(currentThread, contextClassLoader);
+            final ClassLoader ctxCL = contextClassLoader;
+            final ClassLoader old = JBossExecutors.getAndSetContextClassLoader(currentThread, ctxCL);
             try {
                 ((ContextHandler)contextHandler).runWith(delegate, context);
             } catch (Throwable t) {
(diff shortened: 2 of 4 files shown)


src/main/java/org/jboss/threads/JBossExecutors.java
```diff
@@ -336,10 +336,15 @@ static ClassLoader getContextClassLoader(final Thread thread) {
...
-            return getContextClassLoader(thread);
+            final ClassLoader cl = getContextClassLoader(thread);
+            same = (cl == newClassLoader);
+            return cl;
         } finally {
-            setContextClassLoader(thread, newClassLoader);
+            if (!same) {
+                setContextClassLoader(thread, newClassLoader);
+            }
         }
```
carterkozak, line 348: I wonder if we'd be better off removing the try/finally here?

```java
final ClassLoader cl = getContextClassLoader(thread);
if (newClassLoader != newClassLoader) setContextClassLoader(thread, newClassLoader);
return cl;
```

carterkozak:
If memory serves, I refactored this a bit to avoid adding a bunch of extra stack frames, but none of the systems I work on these days do interesting things with classloaders ;-)

I'm very supportive of avoiding unnecessary writes, though!

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -2500,7 +2501,9 @@ public void run() {
                     // nothing else we can safely do here
                 }
             } finally {
-                JBossExecutors.setContextClassLoader(currentThread, old);
+                if (old != ctxCL) {
+                    JBossExecutors.setContextClassLoader(currentThread, old);
+                }
```
carterkozak, line 2506: I suppose there's some risk here that something within the task has set the ContextClassLoader to an entirely new value (e.g. `executor.execute(() -> Thread.currentThread().setContextClassLoad(otherClassLoader))`, which will be leaked into the threads state.
Perhaps `setContextClassLoader` should check the current value before setting a new one instead of doing the check in `getAndSetContextClassLoader`?

It would be entirely reasonable for my use-cases to add a new static option to avoid context classloader checks entirely, but I don't know if that would be generally acceptable.

src/main/java/org/jboss/threads/JBossExecutors.java
franz1981: yeah I believe is better but maybe there' was a reason for it? I have really no idea!

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
franz1981: > Perhaps setContextClassLoader should check the current value before setting a new one instead of doing the check in getAndSetContextClassLoader?

Good point; Yep, that was indeed my concern re changing the original behaviour i.e. it expect on task completion to restore whatever there was before, regardless any previous change.

I think believe performing 2 loads + checks to be better then 1 load + 2 stores

re

> It would be entirely reasonable for my use-cases to add a new static option to avoid context classloader checks entirely, but I don't know if that would be generally acceptable.

I don't know if the quarkus use case would fall into this, really @dmlloyd might know?

src/main/java/org/jboss/threads/JBossExecutors.java
franz1981: I'm still in thinking mode here!

dmlloyd: I'd drop the `try/finally` - it almost certainly adds nothing, but it does add complexity.

src/main/java/org/jboss/threads/JBossExecutors.java
```diff
@@ -350,7 +355,9 @@ static ClassLoader getAndSetContextClassLoader(final Thread thread, final ClassL
      * @param classLoader the new context class loader
      */
     static void setContextClassLoader(final Thread thread, final ClassLoader classLoader) {
-        unsafe.putObject(thread, contextClassLoaderOffs, classLoader);
+        if (unsafe.getObject(thread, contextClassLoaderOffs) != classLoader) {
+            unsafe.putObject(thread, contextClassLoaderOffs, classLoader);
```
dmlloyd, line 359: Too bad there's no `compareAndSetIfDifferent` :)

Do we need to care about the potential race introduced by uncoupled get and set?

franz1981: I hope not, given that's a thread class loader, I hope no others would modify it concurrently, but I'm not an expert on this subject

Sanne: I think the proposed patch has the same semantics as the existing one, in terms of races

src/main/java/org/jboss/threads/ContextClassLoaderSavingRunnable.java
```diff
@@ -12,6 +12,7 @@ class ContextClassLoaderSavingRunnable implements Runnable {
 
     public void run() {
         final Thread currentThread = Thread.currentThread();
+        final ClassLoader newCl = loader;
```
dmlloyd, line 15: Unused variable?

franz1981:
@dmlloyd Adding this to not forget: https://bugs.openjdk.org/browse/JDK-8226731

this likely means that if the Thread where we want to set the class loader and the class loader belong to different regions and are both old gens (they should be likely) the JIT would inject an addtional `StoreLoad` barrier - that would explain why the cost was looking so high (I didn't expected the assembly this time or I would have spotted the `lock addl` )

PR #155, opened 2024-02-14
Inline baseline JDK methods from JDKSpecific

carterkozak:
Now that jboss-threads uses JDK-11+, these methods no longer use a differnt implementation per jdk release.

No strong opinion on this, I just thought I might propose the change after https://github.com/jbossas/jboss-threads/commit/8c87402c7a195c29e6debcad6a8668724cb08cc7


dmlloyd (approved):

dmlloyd:
Makes sense, thanks!

PR #169, opened 2024-05-01
Move unshared fields to arrays

dmlloyd:
The arrays can then be sized based on the run time cache line size, allowing better parallelism on CPUs with long cache lines.


dmlloyd:
@franz1981 WDYT? Is this reasonable?

franz1981:
looking bud @dmlloyd !

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -287,6 +322,31 @@ public final class EnhancedQueueExecutor extends EnhancedQueueExecutorBase6 impl
...
+            int cacheLine = CacheInfo.getSmallestDataCacheLineSize();
+            if (cacheLine == 0) {
+                // guess
+                cacheLine = 64;
+            }
+            int longScale = unsafe.arrayIndexScale(long[].class);
+            int objScale = unsafe.arrayIndexScale(Object[].class);
+            unsharedObjectsSize = cacheLine / objScale * (numUnsharedObjects + 1);
+            unsharedLongsSize = cacheLine / longScale * (numUnsharedLongs + 1);
+            headOffset = unsafe.arrayBaseOffset(Object[].class) + (cacheLine / objScale);
```
franz1981, line 344: Assuming 64 bytes cache line, COOPS and 0 array base offset, it means 

headOffset := 0 + (64 /4 ) = 16
tailOffset := 0 + (64 /4 ) * 2 = 32

which means that if both are within `unsharedObjects`, their distance is just 16 bytes, which is not enough to prevent false-sharing between them

They should be in 2 separate cache lines, if we don't want to interact with each others, but if we expect, in the common scenario that they would be invalidated and used always together, it's ok.

dmlloyd: You're right, I don't know why I did this division. I guess I was thinking of array indexes rather than bytes for some reason.

dmlloyd:
Also, move the queue size field to an unshared memory location since it is also frequently updated.

dmlloyd:
This does need to be tested with a native build on Quarkus before merging.

dmlloyd:
Seems to work! ✅

franz1981:
Sorry for the late approval @dmlloyd but is good, well done!

PR #212, opened 2024-12-30
Don't load Version class unless logging is enabled

geoand:
In Quarkus, version logging is [disabled](https://github.com/quarkusio/quarkus/blob/3.17.5/core/runtime/src/main/java/io/quarkus/runtime/logging/JBossVersion.java#L10) so there is no reason to pay the (admittedly small but not zero) price of loading the class

P.S. I don't see any other side effects that loading the class might have, but I'd welcome more info if there are

### DIFF
--- a/src/main/java/org/jboss/threads/JBossThread.java
+++ b/src/main/java/org/jboss/threads/JBossThread.java
@@ -34,7 +34,9 @@ public class JBossThread extends Thread {
     }).intValue();
 
     static {
-        Version.getVersionString();
+        if (VersionLogging.shouldLogVersion()) {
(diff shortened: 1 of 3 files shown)


geoand:
Exactly correct

dmlloyd (approved):

Sanne:
I love the idea here, but it seems it's not effective as there's a second entry point initializing the Version metadata:

https://github.com/jboss/jboss-threads/blob/main/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java#L87-L90

Should we add the same check?

PR #215, opened 2025-01-07
Remove usages of `Unsafe`

dmlloyd:
Remove all usages of `Unsafe` in favor of supported APIs.

To use the thread-local resetter will now require `--add-opens java.base/java.lang=org.jboss.threads` (or `ALL-UNNAMED` if not using JDK modules).

Java 25+ reports this:
```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by org.jboss.threads.JBossExecutors (jar:file:/Volumes/Case-sensitive/david/.../org/jboss/threads/jboss-threads-3.8.0.Final.jar!/org/jboss/threads/ContextHandler.class)
WARNING: Please consider reporting this to the maintainers of class org.jboss.threads.JBossExecutors
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
```

### DIFF
--- a/src/main/java/org/jboss/threads/JBossExecutors.java
+++ b/src/main/java/org/jboss/threads/JBossExecutors.java
@@ -1,23 +1,18 @@
 package org.jboss.threads;
 
-import java.lang.reflect.Field;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ScheduledExecutorService;
-import java.security.PrivilegedAction;
-import java.security.AccessController;
 
 import org.jboss.logging.Logger;
 import io.smallrye.common.constraint.Assert;
-import sun.misc.Unsafe;
 
 /**
  * JBoss thread- and executor-related utility and factory methods.
  */
-@SuppressWarnings("deprecation")
 public final class JBossExecutors {
 
     private static final Logger THREAD_ERROR_LOGGER = Logger.getLogger("org.jboss.threads.errors");
@@ -225,7 +220,7 @@ public final class JBossExecutors {
 
     private static final Runnable TCCL_RESETTER = new Runnable() {
         public void run() {
-            Thread.currentThread().setContextClassLoader(null);
+            JDKSpecific.setThreadContextClassLoader(Thread.currentThread(), null);
         }
 
         public String toString() {
@@ -294,31 +289,6 @@ public final class JBossExecutors {
         return new ContextClassLoaderSavingRunnable(getContextClassLoader(Thread.currentThread()), delegate);
     }
 
-    static final Unsafe unsafe;
-
-    static final long contextClassLoaderOffs;
-
-    static {
-        unsafe = AccessController.doPrivileged(new PrivilegedAction<Unsafe>() {
-            public Unsafe run() {
-                try {
-                    final Field field = Unsafe.class.getDeclaredField("theUnsafe");
-                    field.setAccessible(true);
-                    return (Unsafe) field.get(null);
-                } catch (IllegalAccessException e) {
-                    throw new IllegalAccessError(e.getMessage());
-                } catch (NoSuchFieldException e) {
-                    throw new NoSuchFieldError(e.getMessage());
-                }
-            }
-        });
-        try {
-            contextClassLoaderOffs = unsafe.objectFieldOffset(Thread.class.getDeclaredField("contextClassLoader"));
-        } catch (NoSuchFieldException e) {
-            throw new NoSuchFieldError(e.getMessage());
-        }
-    }
-
     /**
      * Privileged method to get the context class loader of the given thread.
      *
@@ -326,7 +296,7 @@ public final class JBossExecutors {
      * @return the context class loader
      */
     static ClassLoader getContextClassLoader(final Thread thread) {
-        return (ClassLoader) unsafe.getObject(thread, contextClassLoaderOffs);
+        return JDKSpecific.getThreadContextClassLoader(thread);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/main/java/org/jboss/threads/JBossExecutors.java
```diff
@@ -337,12 +314,21 @@ static ClassLoader getContextClassLoader(final Thread thread) {
      * @return the old context class loader
      */
     static ClassLoader getAndSetContextClassLoader(final Thread thread, final ClassLoader newClassLoader) {
-        final ClassLoader currentClassLoader = (ClassLoader) unsafe.getObject(thread, contextClassLoaderOffs);
-        if (currentClassLoader != newClassLoader) {
-            // not using setContextClassLoader to save loading the current one again
-            unsafe.putObject(thread, contextClassLoaderOffs, newClassLoader);
+        final SecurityManager sm = System.getSecurityManager();
+        if (sm != null) {
+            return AccessController.doPrivileged((PrivilegedAction<ClassLoader>) () -> {
```
franz1981, line 319: this can be a problem..if happen in the hot path :(

src/main/java/org/jboss/threads/ThreadLocalResettingRunnable.java
```diff
@@ -19,22 +25,32 @@ public String toString() {
...
             } catch (NoSuchFieldException e) {
                 throw new NoSuchFieldError(e.getMessage());
             }
         }
 
         static void run() {
             final Thread thread = Thread.currentThread();
-            JBossExecutors.unsafe.putObject(thread, threadLocalMapOffs, null);
-            JBossExecutors.unsafe.putObject(thread, inheritableThreadLocalMapOffs, null);
+            try {
```
franz1981, line 46: We gotta ask @galderz how this perform in native

src/main/java/org/jboss/threads/JBossExecutors.java
dmlloyd: Yeah. I'm thinking maybe we should have an MR JAR overlay here which just calls the target method on 24+.

dmlloyd: The alternative would be a `VarHandle` and require `--add-opens java.base/java.lang=xxx` where `xxx` is `org.jboss.threads` or `ALL-UNNAMED`. Not a good alternative IMO.

Sanne: Can't simply strip out the use of `SecurityManager` yet?

dmlloyd: IIRC WildFly will support the security manager until it no longer supports Jakarta EE 10, which is still in the future AFAIK. That said, we could probably do something like have a Java 24+ MR-JAR layer that skips it, perhaps, and use Unsafe on the lower layers?

Sanne: FWIW, I had stripped all Hibernate libraries from any reference to the SecurityManager - but this applies only to next-gen major versions, so consumers have the option to use the older one.

jamezp: WildFly and EAP need to continue to support the security manager. Especially for Jakarta EE 10 as it's required per the spec. Ripping out privileged blocks could potentially require WildFly to fork projects or stick with older versions, neither of which is ideal.

That said, we can't control what the JDK does. If we reach a point where the JDK removes the API's, then we cannot support that version of the JVM. However, until then we prefer projects keep privileged blocks in tact.

Sanne: Thanks @jamezp for intervening! Yet we could have such a change in a minor (or even major?) and only update selected versions of future Wildfly for which this is fine? I'd do that over MR, also because all such optimisations being discussed here don't quite belong in the long-term-maintained branches.

jamezp: TBH it's really up to the project. We need to continue support for the security manager for various reasons. If a project WildFly/EAP consumes removes it and it's a Red Hat/JBoss project, there is a good chance you'll have to support multiple branches/releases :)

What I mean is, it really comes down to what the project wants to do. Supporting two different release streams is fine, if the project is okay with that.

For the projects I lead, I just leave the security manger code because I personally don't want to have to support multiple release streams for who knows how long :)

src/main/java/org/jboss/threads/ThreadLocalResettingRunnable.java
dmlloyd: I expect it to be OK: AFAIK, VarHandle is using these method handles internally, and we're acquiring them in static init. My expectation is that the calls will be inlined.

However, either way it's not going to work without `--add-opens java.base/java.lang=org.jboss.threads` or `...=ALL-UNNAMED`. That's the real sticky part.

src/main/java/org/jboss/threads/JBossExecutors.java
dmlloyd: My easy fix for this is to use `Unsafe` up to JDK 23, and then use the normal accessors for 24+, since in 24+ there are no security manager checks and thus the abstraction does not do anything special anymore.

dmlloyd:
This now just needs:

* Java 24 release to be finalized
* GitHub runners to support Java 24

dmlloyd:
Tests pass with `24-ea`, but we might as well wait a couple weeks longer for 24 to be finalized.

geoand:
@dmlloyd @cescoffier if we want to use JDK 24 with Quarkus LTS, this will be needed to avoid 

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by org.jboss.threads.JBossExecutors (file:/opt/homebrew/Cellar/quarkus/3.21.0/libexec/lib/quarkus-cli-3.21.0-runner.jar)
WARNING: Please consider reporting this to the maintainers of class org.jboss.threads.JBossExecutors
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
```

cc @danieloh who ran into this

PR #216, opened 2025-01-08
Save complex OS-level I/O to know the CPU topology

franz1981:
This is related https://github.com/smallrye/smallrye-common/issues/370 since `smallrye-common` exposes `CacheInfo` as a public API and dropping it can impact existing users - while doing it here, won't impact anyone. 

The benefits are https://github.com/quarkusio/quarkus/issues/45436

### DIFF
--- a/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
+++ b/src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
@@ -47,7 +47,6 @@ import java.util.concurrent.locks.ReentrantLock;
 import javax.management.ObjectInstance;
 import javax.management.ObjectName;
 
-import io.smallrye.common.cpu.CacheInfo;
 import org.jboss.threads.management.ManageableThreadPoolExecutorService;
 import org.jboss.threads.management.StandardThreadPoolMXBean;
 
@@ -122,13 +121,6 @@ public final class EnhancedQueueExecutor extends AbstractExecutorService impleme
     // =======================================================
     // Optimization control flags
     // =======================================================
-
-    /**
-     * A global hint which establishes whether it is recommended to disable uses of {@code EnhancedQueueExecutor}.
-     * This hint defaults to {@code false} but can be changed to {@code true} by setting the {@code jboss.threads.eqe.disable}
-     * property to {@code true} before this class is initialized.
-     */
-    public static final boolean DISABLE_HINT = readBooleanPropertyPrefixed("disable", false);
 
     /**
      * Update the summary statistics.
@@ -337,13 +329,10 @@ public final class EnhancedQueueExecutor extends AbstractExecutorService impleme
         private static final long queueSizeOffset;
 
         static {
-            int cacheLine = CacheInfo.getSmallestDataCacheLineSize();
-            if (cacheLine == 0) {
-                // guess
-                cacheLine = 64;
-            }
+            // this is fine for pretty much 32 and 64 bit x86 and ARM processors; see
+            // https://github.com/ziglang/zig/blob/0.13.0/lib/std/atomic.zig#L424-L434
             // cpu spatial prefetcher can drag 2 cache-lines at once into L2
-            int pad = cacheLine > 128 ? cacheLine : 128;
+            int pad = 128;
             int longScale = unsafe.arrayIndexScale(long[].class);
             int taskNodeScale = unsafe.arrayIndexScale(TaskNode[].class);
             // these fields are in units of array scale


src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -337,13 +329,10 @@ private static final class RuntimeFields {
...
-            int cacheLine = CacheInfo.getSmallestDataCacheLineSize();
-            if (cacheLine == 0) {
-                // guess
-                cacheLine = 64;
-            }
+            // this is fine for pretty much 32 and 64 bit x86 and ARM processors; see
+            // https://github.com/ziglang/zig/blob/0.13.0/lib/std/atomic.zig#L424-L434
             // cpu spatial prefetcher can drag 2 cache-lines at once into L2
-            int pad = cacheLine > 128 ? cacheLine : 128;
+            int pad = 128;
```
franz1981, line 335: @dmlloyd I think I can drop the pad (or set it to 0) in case the number of processors is 1 - which CAN happen with containers, wdyt?  Maybe not that worthy, but still...

dmlloyd: Sure, as long as we can be sure that `1` means `1` and not e.g. `2×½`.

franz1981: thanks, yep, I'll make sure - I still feel bad about my bug on jboss logger encoding :"(

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -123,13 +122,6 @@ public final class EnhancedQueueExecutor extends AbstractExecutorService impleme
     // Optimization control flags
     // =======================================================
 
-    /**
-     * A global hint which establishes whether it is recommended to disable uses of {@code EnhancedQueueExecutor}.
-     * This hint defaults to {@code false} but can be changed to {@code true} by setting the {@code jboss.threads.eqe.disable}
-     * property to {@code true} before this class is initialized.
-     */
-    public static final boolean DISABLE_HINT = readBooleanPropertyPrefixed("disable", false);
-
```
dmlloyd, line 132: This is part of the API, so removing it would be a problem. We could deprecate it and set it to `false` instead though.

franz1981: let me remove from this PR than, better track it separately

src/main/java/org/jboss/threads/EnhancedQueueExecutor.java
```diff
@@ -337,23 +336,31 @@ private static final class RuntimeFields {
...
+                // we both pad before and after the array to avoid false sharing with surrounding heap objects
+                unsharedTaskNodesSize = pad / taskNodeScale * (numUnsharedObjects + 1);
+                unsharedLongsSize = pad / longScale * (numUnsharedLongs + 1);
+                // these fields are in bytes
+                headOffset = unsafe.arrayBaseOffset(TaskNode[].class) + pad;
+                tailOffset = unsafe.arrayBaseOffset(TaskNode[].class) + pad * 2;
+                threadStatusOffset = unsafe.arrayBaseOffset(long[].class) + pad;
+                queueSizeOffset = unsafe.arrayBaseOffset(long[].class) + pad * 2;
+            } else {
+                unsharedTaskNodesSize = numUnsharedObjects;
```
franz1981, line 355: @dmlloyd we have no coverage for this unless we run our test suite with `-XX:ActiveProcessorCount=1`
so please take a look if I did something wrong

dmlloyd:
This PR will conflict with #215 so we should coordinate somehow.

franz1981:
since is a small change @dmlloyd here, feel free to cp into your branch without authoring me :heart:

Sanne:
Could we take it a step further?
 - https://github.com/jbossas/jboss-threads/commit/3de2a0a155596d44b54516e2d4aeb6af3df80426

geoand:
I vote for having a sane default and having people opt-in to more complex handling as:

* The default is probably fine for almost all cases we are trying to address
* Trying to figure out the actual cache line info is one of the few things whose elimination has an undeniably measurable impact (especially on Mac)

franz1981:
@dmlloyd do you think we can make this one to slip in first? So we can take our time into verifying that Unsafe-less we don't regress in runtime perf too - not just RSS? :pray:

dmlloyd:
I think we should fix this only in `smallrye-common`. Otherwise, as soon as someone else uses those functions, we'll pay the cost anyway, and we'll end up with two different analyses of cache lines.

geoand:
https://github.com/smallrye/smallrye-common/pull/389 is a simply way to do that

